PR #18, opened 2020-11-05
Sasha/delete model run

sasha-scale:
Added new delete model run functionality into library.  Is pretty simple, tested by deleting some of my extraneous runs on personal datasets.


nucleus/__init__.py
```diff
@@ -86,6 +86,8 @@ class NucleusClient:
     """
 
     def __init__(self, api_key: str):
+        print("HERE")
+        print(NUCLEUS_ENDPOINT)
```
Nastia39, line 90: delete these lines

Nastia39 (changes requested):

nucleus/__init__.py
sasha-scale: oof sorry about that, resolved.

Nastia39 (approved):

PR #23, opened 2020-12-15
removed session from both local + remote upload

srikanth-scale:
Removed the session object because of SSL Handshake issue


sasha-scale (approved):
LGTM

PR #26, opened 2021-02-05
Append to slice api

jihan-yin:
Adds function to slice instance that lets users append new images to the slice. Tested locally, working with conjunction with https://github.com/scaleapi/scaleapi/pull/21392.

### DIFF
--- a/nucleus/slice.py
+++ b/nucleus/slice.py
@@ -49,7 +49,7 @@ class Slice:
         """
         if dataset_item_ids and reference_ids:
             raise Exception(
-                "You cannot both dataset_item_ids and reference_ids"
+                "You cannot specify both dataset_item_ids and reference_ids"
             )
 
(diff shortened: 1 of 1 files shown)


nucleus/slice.py
```diff
@@ -27,3 +27,38 @@ def info(self, id_type: str = DATASET_ITEM_ID_KEY) -> dict:
...
+        reference_ids: List[str],
+
+        :return:
+        {
+            "slice_id": str,
+        }
+        """
+        if dataset_item_ids and reference_ids:
+            raise Exception(
+                "You cannot both dataset_item_ids and reference_ids"
```
sasha-scale, line 52: grammar error with message here: "You cannot specify both dataset_item_ids and reference_ids"

sasha-scale (approved):
one small nit, otherwise looks good (tested locally).  Merge scaleapi PR first!

PR #28, opened 2021-02-08
added tqdm as dependancy

srikanth-scale:


rkaplan:
`requirements.txt` is autogenerated with pigar in our current workflow, let's re-autogenerate rather than add manually

rkaplan (approved):

PR #30, opened 2021-02-11
Annotation Idempotency

jihan-yin:
Adds `annotation_id` as optional parameter to annotations and prediction classes. This will be used to uniquely identify each annotation/prediction within its dataset item. Also adds `update` as an optional boolean parameter when uploading predictions and annotations. Setting `update=False`, the default behavior, will skip uploading annotations to dataset images that already have annotations with the same `annotation_id`. Setting `update=True` will overwrite such annotations.


nucleus/__init__.py
```diff
@@ -128,7 +128,7 @@ def __init__(self, api_key: str, use_notebook: bool = False):
         self.api_key = api_key
         self.tqdm_bar = tqdm.tqdm
         if use_notebook:
-            self.tqdm_bar = tqdm.notebook.tqdm
+            self.tqdm_bar = tqdm_notebook.tqdm
```
jihan-yin, line 131: Was running into issues calling the api from within a notebook, this resolved it for me

nucleus/__init__.py
```diff
@@ -462,13 +462,15 @@ def exception_handler(request, exception):
     def annotate_dataset(
         self,
         dataset_id: str,
-        annotations: List[DatasetItem],
+        annotations: List[Union[BoxAnnotation, PolygonAnnotation]],
```
sasha-scale, line 465: omg - nice fix

nucleus/__init__.py
```diff
@@ -462,13 +462,15 @@ def exception_handler(request, exception):
...
+        update: bool,
```
sasha-scale, line 466: see comment on sister PR - I think this should be optional with default value False

nucleus/__init__.py
```diff
@@ -602,7 +600,11 @@ def predict(
         tqdm_batches = self.tqdm_bar(batches)
 
         for batch in tqdm_batches:
-            batch_payload = {ANNOTATIONS_KEY: batch}
+            batch_payload = construct_box_predictions_payload(
+                annotations,
+                update,
+            )
+            print(batch_payload)
```
sasha-scale, line 607: remove print statement before merging

nucleus/constants.py
```diff
@@ -1,4 +1,5 @@
-NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
+#NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
```
sasha-scale, line 1: remove before merging

nucleus/dataset.py
```diff
@@ -10,11 +10,14 @@
     REFERENCE_IDS_KEY,
     NAME_KEY,
     ITEM_KEY,
+    DEFAULT_ANNOTATION_UPDATE_MODE,
     ANNOTATIONS_KEY,
 )
 from .payload_constructor import construct_model_run_creation_payload
 
 
+
```
sasha-scale, line 19: :nit: extraneous white space

sasha-scale (changes requested):
I was unable to upload GT annotations without specifying the annotation_id, the error message was "annotation_id must be a string."  I think the issue was that when I did not specify it, it got sent as `None` in the request payload which didn't make it through schema validation

sasha-scale (approved):
LGTM, tested locally.  Response format does not have annotations_ignored which to me blocks merge, but that is more a comment on the scaleapi Pr than this one.

The unit tests are incredibly thorough!  great jobs!

tests/helpers.py
```diff
@@ -0,0 +1,90 @@
...
+    assert annotation_instance.label == annotation_dict['label']
+    assert annotation_instance.annotation_id == annotation_dict['annotation_id']
+    for instance_pt, dict_pt in zip(annotation_instance.vertices, annotation_dict['vertices']):
+        assert instance_pt['x'] == dict_pt['x']
+        assert instance_pt['y'] == dict_pt['y']
+
+# Asserts that a box prediction instance matches a dict representing its properties.
+# Useful to check prediction uploads/updates match.
+def assert_box_prediction_matches_dict(prediction_instance, prediction_dict):
+    assert_box_annotation_matches_dict(prediction_instance, prediction_dict)
```
rkaplan, line 85: Infinite recursion?

tests/helpers.py
```diff
@@ -0,0 +1,90 @@
...
+    assert prediction_instance.confidence == prediction_dict['confidence']
+
+def assert_polygon_prediction_matches_dict(prediction_instance, prediction_dict):
+    assert_polygon_annotation_matches_dict(prediction_instance, prediction_dict)
```
rkaplan, line 89: Infinite recursion?

tests/helpers.py
jihan-yin: It's a different function 😄

tests/helpers.py
jihan-yin: same as above

rkaplan (approved):

PR #31, opened 2021-02-12
Sasha/list autotags

sasha-scale:
Add method on Dataset class to list autotag ids for dataset.  In follow up PR will define Autotag class with info and list_items member methods, but that needs to be preceeded by REST API implementation


tests/test_dataset.py
```diff
@@ -83,3 +83,11 @@ def check_is_expected_response(response):
 
     response = dataset.append(ds_items_with_metadata)
     check_is_expected_response(response)
+
+def test_dataset_list_autotags(CLIENT):
+    # Creation
+    ds = CLIENT.create_dataset(TEST_DATASET_NAME)
```
rkaplan, line 89: Have to delete the dataset after. Should just use the dataset fixture here

rkaplan (approved):
LGTM besides changing the unit test to use a fixture

PR #32, opened 2021-02-12
Pytests for Slice creation and append

jihan-yin:
Added pytests for testing slice creation and append. Also added some minor type and grammar fixes.

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -14,6 +14,7 @@ from nucleus.constants import (
 
 
 TEST_DATASET_NAME = '[PyTest] Test Dataset'
+TEST_SLICE_NAME = '[PyTest] Test Slice'
 TEST_IMG_URLS = [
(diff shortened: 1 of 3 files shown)


tests/test_dataset.py
```diff
@@ -22,14 +23,13 @@
     's3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/89b42832-10d662f4.jpg',
 ]
 
-@pytest.fixture(scope='module')
+@pytest.fixture()
```
jihan-yin, line 26: Decided to do this so our different test functions have separate state. Otherwise, if one function causes the dataset to bug out somehow, everything else will fail from that.

sasha-scale: good catch!

sasha-scale (approved):
LGTM thanks for the copy fixes :)

PR #33, opened 2021-02-13
add function to delete models

sasha-scale:
Add support for deleting models


jihan-yin (approved):
Tested with new endpoint, lgtm

PR #34, opened 2021-02-17
added ability to list models

srikanth-scale:
Adds support for listing models, and printing the model to console in a human readable way.

Tested locally, with and without models being present for a given user.


nucleus/model.py
```diff
@@ -24,6 +24,9 @@ def __init__(
         self.metadata = metadata
         self._client = client
 
+    def __repr__(self):
+        return f'Model({self.id}, {self.name}, {self.reference_id}, {self.metadata}, {self._client})'
```
rkaplan, line 28: __repr__ implementations should include argument names as kwargs so that the output results are readable.

```python
return f'Model(model_id={self.id}, name={self.name}, reference_id={self.reference_id}, metadata={self.metadata}, client={self._client})'
```

nucleus/__init__.py
```diff
@@ -131,13 +131,18 @@ def __init__(self, api_key: str, use_notebook: bool = False):
...
         :return: model_ids
         """
-        # TODO implement API
-        raise NotImplementedError
+        model_objects = self._make_request({}, "models/", requests.get)
+
+        try:
+            models = [Model(model['id'], model["name"], model["ref_id"], model["metadata"], self) for model in model_objects['models']]
+            return models
+        except e:
```
rkaplan, line 144: Why do we have a blanket try/except here?

nucleus/__init__.py
```diff
@@ -131,13 +131,18 @@ def __init__(self, api_key: str, use_notebook: bool = False):
...
         """
         Lists available models in your repo.
         :return: model_ids
         """
-        # TODO implement API
-        raise NotImplementedError
+        model_objects = self._make_request({}, "models/", requests.get)
+
+        try:
+            models = [Model(model['id'], model["name"], model["ref_id"], model["metadata"], self) for model in model_objects['models']]
```
rkaplan, line 142: nit: quotation marks are inconsistent

nucleus/model.py
```diff
@@ -24,6 +24,9 @@ def __init__(
         self.metadata = metadata
         self._client = client
 
+    def __repr__(self):
```
rkaplan, line 27: missing return type annotation

nucleus/__init__.py
srikanth-scale: catching errors if json is badly formatted/ no models exist returns empty array

rkaplan: If no models exist (i.e. `model_objects['models']` is an empty array), then the list comprehension will be an empty array, it won't crash, so this is unnecessary. 

If the JSON for one model is badly formatted, we should not silently repress all results for all models. The error should be loud and noisy—this means there is a server-side bug on our end.

In general, there is almost never a good reason to have a blanket except statement that catches every possible type of exception. Should avoid this antipattern.

rkaplan:
Left comments. Please also add a unit test for the new endpoint.

tests/test_models.py
```diff
@@ -0,0 +1,50 @@
...
+from nucleus.constants import (
+    NEW_ITEMS,
+    UPDATED_ITEMS,
+    IGNORED_ITEMS,
+    ERROR_ITEMS,
+    ERROR_PAYLOAD,
+    DATASET_ID_KEY,
+)
+
+TEST_MODEL_NAME = '[PyTest] Test Model'
```
rkaplan, line 13: Most of these should be imported from `https://github.com/scaleapi/nucleus-python-client/blob/master/tests/helpers.py`

tests/test_models.py
```diff
@@ -0,0 +1,50 @@
...
+TEST_REFERENCE_ID = '[PyTest] Test Model'
+TEST_METADATA = {
+    'key': 'value'
+}
+TEST_MODEL_RUN_NAME = '[PyTest] Test ModelRun'
+TEST_DATASET_NAME = '[PyTest] Test Dataset'
+TEST_SLICE_NAME = '[PyTest] Test Slice'
```
rkaplan, line 20: Unused

tests/test_models.py
```diff
@@ -0,0 +1,50 @@
...
+TEST_IMG_URLS = [
+    "s3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/6dd63871-831611a6.jpg",
+    "s3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/82c1005c-e2d1d94f.jpg",
+    "s3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/7f2e1814-6591087d.jpg",
+    "s3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/06924f46-1708b96f.jpg",
+    "s3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/89b42832-10d662f4.jpg",
+]
+
+@pytest.fixture()
+def dataset(CLIENT):
```
rkaplan, line 30: Rather than copy/pasting this fixture from `test/test_dataset.py`, it should be just moved to `conftest.py` and used from both of those other files

tests/test_models.py
```diff
@@ -0,0 +1,50 @@
...
+    assert isinstance(m, Model)
+    assert isinstance(m_run, ModelRun)
+
+    # List
+    ms = CLIENT.list_models()
+
+    assert m in ms
+
+    CLIENT.delete_model(m.id)
+    CLIENT.delete_model_run(m_run.model_run_id)
```
rkaplan, line 50: should test listing models again after deletion and ensure the empty list is returned

tests/test_models.py
```diff
@@ -0,0 +1,50 @@
...
+
+TEST_MODEL_NAME = '[PyTest] Test Model'
+TEST_REFERENCE_ID = '[PyTest] Test Model'
+TEST_METADATA = {
+    'key': 'value'
+}
+TEST_MODEL_RUN_NAME = '[PyTest] Test ModelRun'
+TEST_DATASET_NAME = '[PyTest] Test Dataset'
+TEST_SLICE_NAME = '[PyTest] Test Slice'
+TEST_IMG_URLS = [
```
rkaplan, line 21: Unused

tests/test_models.py
```diff
@@ -0,0 +1,50 @@
...
+    m = CLIENT.add_model(TEST_MODEL_NAME, TEST_REFERENCE_ID, TEST_METADATA)
+    m_run =  m.create_run(TEST_MODEL_RUN_NAME, TEST_DATASET, TEST_PREDS, TEST_METADATA)
+
+    assert isinstance(m, Model)
+    assert isinstance(m_run, ModelRun)
+
+    # List
+    ms = CLIENT.list_models()
+
+    assert m in ms
```
rkaplan, line 47: Better to assert exactly what the model list is expected to be

srikanth-scale: But then what if it contains models that I've not created in the context of this test? The API key belongs to me and I create models randomly, no way for me to know them beforehand

tests/test_models.py
srikanth-scale: moved

tests/test_models.py
srikanth-scale: mo ved

tests/test_models.py
rkaplan: Perhaps then you should call `.list_models()` before model creation, save the returned value of the array, then list after model creation, verify the array is the same except for the new addition, then list after model deletion to verify that the array is back to it was as it started?

rkaplan (approved):

PR #35, opened 2021-02-19
Sasha/segmentation

sasha-scale:
Add support for semseg annotation upload via python client.  Now, users can use the `dataset.annotate` method to upload any of the three supported annotation types.  Under the hood, we separate segmentation annotations from the rest in order to make a request to a separate endpoint.

Test Cases are:
(1). User can upload single semseg annotation
(2). User can upload batch of semseg annotations
(3). Not specifying `force` flag results in ignoring conflicts
(4). Specifying `force = True` results in update on conflict
(5). User can upload mixed batch of annotations, i.e. 5 bboxs + 5 segmentation masks = 10 successfully processed annotations

Uploading predictions will be a separate PR.  Blocking merge: change force flag to be called `update` everywhere, and add python documentation for Segmentation.


tests/helpers.py
```diff
@@ -1,90 +1,142 @@
...
-    's3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/89b42832-10d662f4.jpg',
+    "s3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/6dd63871-831611a6.jpg",
+    "s3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/82c1005c-e2d1d94f.jpg",
+    "s3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/7f2e1814-6591087d.jpg",
+    "s3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/06924f46-1708b96f.jpg",
+    "s3://scaleapi-attachments/BDD/BDD/bdd100k/images/100k/train/89b42832-10d662f4.jpg",
 ]
 
+
+def get_signed_url(url):
```
sasha-scale, line 21: adding these helper functions was the best way I could think of to make testing with semseg masks feasible (need to sign the URL before upload).  It's a bit clunky, and I'm open to suggestions for how to avoid signed urls in the future

tests/test_annotation.py
```diff
@@ -35,118 +44,218 @@ def test_box_gt_upload(dataset):
...
+    assert response["annotations_processed"] == 10
+    assert response["annotations_ignored"] == 0
 
 
 def test_box_gt_upload_update(dataset):
     annotation = BoxAnnotation(**TEST_BOX_ANNOTATIONS[0])
     response = dataset.annotate(annotations=[annotation])
 
-    assert response['annotations_processed'] == 1
+    assert response["annotations_processed"] == 1
```
sasha-scale, line 151: I think the black pre-commit hook standardizes all quotations to " instead of '

nucleus/annotation.py
```diff
@@ -15,9 +15,86 @@
...
+        return payload
+
+
+class SegmentationAnnotation:
+    def __init__(
+        self,
+        mask_url: str,
+        annotations: List[Segment],
+        reference_id: Optional[str] = None,
+        item_id: Optional[str] = None,
```
rkaplan, line 60: Looks like this API is missing support for an `annotation_id`, which we should be consistent about across annotation types

nucleus/payload_constructor.py
```diff
@@ -39,6 +44,17 @@ def construct_annotation_payload(
...
 
+def construct_segmentation_payload(
+    annotation_items: List[SegmentationAnnotation],
+    update: bool,
+) -> dict:
+    annotations = []
+    for annotation_item in annotation_items:
+        annotations.append(annotation_item.to_payload())
+
+    return {SEGMENTATIONS_KEY: annotations, "force": update}
```
rkaplan, line 55: Believe we decided to standardize on `update` as the key name for this parameter instead of `force`

tests/helpers.py
rkaplan: This works fine, though one option in the future is to just move some semseg masks to a public s3 bucket for testing purposes

tests/test_annotation.py
```diff
@@ -35,118 +44,218 @@ def test_box_gt_upload(dataset):
...
+
+def test_single_semseg_gt_upload(dataset):
+    annotation = SegmentationAnnotation.from_json(
+        TEST_SEGMENTATION_ANNOTATIONS[0]
+    )
+    response = dataset.annotate(annotations=[annotation])
+    assert response["dataset_id"] == dataset.id
+    assert response["annotations_processed"] == 1
+    assert response["annotations_ignored"] == 0
+    # assert_box_annotation_matches_dict(response_annotation, TEST_BOX_ANNOTATIONS[0])
```
rkaplan, line 83: In investigating this commented out piece of code, I noticed that running

`response = dataset.refloc(annotation.reference_id)["annotations"]`

Does not return semseg annotations. For a dataset item with both semseg and bbox annotations it only returns the bbox ones. We have to update those backend dataset.refloc/iloc/etc methods to also return semseg annotations as part of full semseg support.

rkaplan: (And make sure it works for polygon annotations, I haven't tested that)

tests/test_annotation.py
```diff
@@ -35,118 +44,218 @@ def test_box_gt_upload(dataset):
...
+    annotations = [
+        SegmentationAnnotation.from_json(ann)
+        for ann in TEST_SEGMENTATION_ANNOTATIONS
+    ]
+    response = dataset.annotate(annotations=annotations)
+    assert response["dataset_id"] == dataset.id
+    assert response["annotations_processed"] == 5
+    assert response["annotations_ignored"] == 0
+
+    # When we re-upload, expect them to be ignored
```
rkaplan, line 126: Nit, comment needs updating

nucleus/annotation.py
rkaplan: (I know there isn't an idempotency benefit in this case because there can only be one semseg mask per image, unlike bounding boxes, but I think it still makes sense to get users in the habit of always providing annotation_ids with their annotations and have consistency across our API. We will want to support querying by annotation_id directly in the query bar down the line, e.g., which would be relevant for semseg masks too. Also useful for when we eventually support GT versioning.)

tests/test_annotation.py
sasha-scale: yep this change should be preceded by a PR in the scaleapi repo but should be pretty quick!

nucleus/annotation.py
sasha-scale: I would recommend saving the annotation_id for a separate follow-up PR because the changes required are mostly on the scaleapi side.  I imagine since uniqueness is already enforced on dataset items, the main value prop is annotation_ids would not allow you upload same mask to two different items

nucleus/payload_constructor.py
sasha-scale: yup, changed it here now that the switch is deployed on prod

tests/helpers.py
sasha-scale: I'll ask the platform team if we're allowed to make public buckets

nucleus/annotation.py
rkaplan: Sgtm

rkaplan (approved):
LGTM

PR #38, opened 2021-02-26
Sasha/fix bugs

sasha-scale:
after my latest PR merge I noticed that pytest was broken for me for some reason, changing the import statement `from helpers import *` to a relative import fixed it for me, curious if anyone else has the same issue.

### DIFF
--- a/nucleus/constants.py
+++ b/nucleus/constants.py
@@ -1,5 +1,4 @@
-# NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
-NUCLEUS_ENDPOINT = "http://localhost:3000/v1/nucleus"
+NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
 ITEMS_KEY = "items"
 ITEM_KEY = "item"
 REFERENCE_ID_KEY = "reference_id"
--- a/nucleus/model.py
+++ b/nucleus/model.py
@@ -1,6 +1,10 @@
 from typing import List, Optional, Dict, Union
 from .dataset import Dataset
-from .prediction import BoxPrediction, PolygonPrediction
+from .prediction import (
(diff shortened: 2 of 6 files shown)


jihan-yin:
Looks like the issue is from https://github.com/scaleapi/nucleus-python-client/commit/560d9c6c1f9f5c02554575c301e1c9aee231a7bc, @srikanth-scale added an `init.py` to `tests/` to make it a module, so we can no longer treat its files as individual packages. This is good practice so not really an issue, thanks for fixing the imports though!

nucleus/constants.py
```diff
@@ -1,5 +1,4 @@
-# NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
-NUCLEUS_ENDPOINT = "http://localhost:3000/v1/nucleus"
+NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
```
jihan-yin, line 1: 🙈

nucleus/model.py
```diff
@@ -1,6 +1,10 @@
 from typing import List, Optional, Dict, Union
 from .dataset import Dataset
-from .prediction import BoxPrediction, PolygonPrediction
+from .prediction import (
+    BoxPrediction,
+    PolygonPrediction,
+    SegmentationPrediction,
```
jihan-yin, line 6: Curious how things were running before we added this import

jihan-yin (approved):
Lgtm, thanks for the fixes!

nucleus/constants.py
sasha-scale: RIGHT

nucleus/model.py
sasha-scale: they weren't...

PR #39, opened 2021-02-26
adjust refloc payload format

sasha-scale:
In the scaleapi PR:
https://github.com/scaleapi/scaleapi/pull/22013

Changed the refloc/iloc/loc payload format from:

```
{
    item: {original_image_url, reference_id}
    annotations: [{id, type, label, geometry}, ... ]
}
```

to

```
{
    item: {original_image_url, reference_id}
    annotations: {
        'box': [{id, type, label, geometry}, ... ],
        'polygon': [{id, type, label, geometry}, ... ],
        'segmentation': {id, mask_url, dataset_item_id, annotations: [...]}
    }
}

```
which better generalizes to the multiple annotation type model.  This PR incorporates this interface change into the nucleus client,  we will merge this one when the other is deployed. To test, check out the corresponding PR in scaleapi (linked above) and change the nucleus client to make requests to the local server.  Then run pytest.


nucleus/constants.py
```diff
@@ -1,5 +1,4 @@
-# NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
-NUCLEUS_ENDPOINT = "http://localhost:3000/v1/nucleus"
+NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
```
rkaplan, line 1: 😬 Have to be careful to not edit this in master, glad this PR is fixing.

We should probably use an environment variable to let python client users override the default endpoint URL, rather than always changing this constant on local branches for testing

nucleus/annotation.py
```diff
@@ -74,7 +74,7 @@ def __str__(self):
     @classmethod
     def from_json(cls, payload: dict):
         return cls(
-            mask_url=payload[MASK_URL_KEY],
+            mask_url=payload.get(MASK_URL_KEY, None),
```
rkaplan, line 77: I don't understand, how can you construct a SegmentationAnnotation without a mask_url in the current implementation of `SegmentationAnnotation`? For one, `mask_url`'s type annotation is `str`, so we cannot pass `None` here. And more broadly I don't see how you can have a `SegmentationAnnotation` without a mask. Should be raising an error if MASK_URL_KEY is not in the payload instead?

nucleus/dataset.py
```diff
@@ -170,6 +178,7 @@ def refloc(self, reference_id: str) -> dict:
         }
         """
         response = self._client.dataitem_ref_id(self.id, reference_id)
+        print("RESPONSE" + str(response))
```
rkaplan, line 181: Remove

nucleus/dataset.py
```diff
@@ -227,17 +236,29 @@ def list_autotags(self):
...
             return response
-        annotations = [
-            BoxAnnotation.from_json(ann)
-            if ann["type"] == "box"
-            else PolygonAnnotation.from_json(ann)
-            for ann in annotation_payload
-        ]
+
+        annotation_response = {}
+        if BOX_TYPE in annotation_payload:
```
rkaplan, line 245: Could consider an outer for loop like

```python
for (type_key, type_cls) in zip([BOX_TYPE, POLYGON_TYPE, SEGMENTATION_TYPE],
        [BoxAnnotation, PolygonAnnotation, SegmentationAnnotation]):
```

to avoid the copypaste if desired, though I'm not sure that's much better.

Best would be if we added a base `Annotation` parent class in annotation.py which implemented the `.from_json` class method itself (which routed to the appropriate subclass's `.from_json` depending on the value of the type key in the json) to return the correct subclass instance. And added a `ANNOTATION_TYPES=(BOX_TYPE, POLYGON_TYPE, SEGMENTATION_TYPE)` tuple to constants.py. Then this implementation could look like just:

```python
for type in ANNOTATION_TYPES:
    if type in annotation_payload:
        annotation_response[type] = [Annotation.from_json(ann) for ann in annotation_payload[type]]
```

nucleus/model_run.py
```diff
@@ -122,12 +127,25 @@ def _format_prediction_response(
...
         annotations = response.get(ANNOTATIONS_KEY, None)
+        print(annotations)
         if annotations:
-            return [
-                BoxPrediction.from_json(ann)
-                if ann["type"] == "box"
-                else PolygonPrediction.from_json(ann)
-                for ann in annotations
-            ]
+            annotation_response = {}
```
rkaplan, line 132: Could also clean this up if we make the above suggested change

nucleus/model_run.py
```diff
@@ -122,12 +127,25 @@ def _format_prediction_response(
         self, response: dict
     ) -> Union[dict, List[Union[BoxPrediction, PolygonPrediction]]]:
         annotations = response.get(ANNOTATIONS_KEY, None)
+        print(annotations)
```
rkaplan, line 130: Remove

tests/helpers.py
```diff
@@ -147,6 +147,16 @@ def assert_polygon_annotation_matches_dict(
         assert instance_pt["y"] == dict_pt["y"]
 
 
+def assert_segmentation_annotation_matches_dict(
+    annotation_instance, annotation_dict
+):
+    assert annotation_instance.mask_url == annotation_dict["mask_url"]
+    # Cannot guarantee segments are in same order
```
rkaplan, line 154: Could consider making segments comparable and then sorting them, and checking the sorted lists are equal

tests/test_prediction.py
```diff
@@ -228,3 +236,5 @@ def test_mixed_pred_upload(model_run):
     assert response["model_run_id"] == model_run.model_run_id
     assert response["predictions_processed"] == 3
     assert response["predictions_ignored"] == 0
+
+    response_refloc = model_run.refloc(prediction_polygon.reference_id)
```
rkaplan, line 240: Are there supposed to be tests on this object as well

nucleus/model_run.py
sasha-scale: I wasn't able to use the base class approach for the predictions case because there's currently no way to differentiate a segmentation_prediction from a segmentation_gt_annotation from the payloads.  can fix this in the future but don't think it should block this PR.

tests/helpers.py
sasha-scale: done, great suggestion!

tests/test_prediction.py
sasha-scale: yes thanks for catching

nucleus/model_run.py
rkaplan: Sgtm

rkaplan (approved):

PR #42, opened 2021-03-15
Sasha/adjust params

sasha-scale:
Increase batch size for model predictions, add timeout to network requests.

While trying to upload a large dataset, I found that my upload script would occasionally hang indefinitely, and I would have to quit the program and re-run.  I think this happened because without a timeout parameter, the request library will keep the connection open forever even if it hears no response from the server.  In the case of a poor internet connection, this leads to hanging.  In my opinion, better for upload to fail due to timeout than to hang!

### DIFF
--- a/nucleus/constants.py
+++ b/nucleus/constants.py
@@ -1,4 +1,5 @@
 NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
+DEFAULT_TIMEOUT = 30
 ITEMS_KEY = "items"
 ITEM_KEY = "item"
 REFERENCE_ID_KEY = "reference_id"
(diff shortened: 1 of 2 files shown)


nucleus/constants.py
```diff
@@ -1,4 +1,5 @@
 NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
+DEFAULT_TIMEOUT = 60
```
rkaplan, line 2: Nit: can you put the units in the name for constants where the unit is ambiguous. E.g. `DEFAULT_TIMEOUT_SEC`

rkaplan (approved):
Lgtm barring nit

PR #43, opened 2021-03-22
Support Custom Index

sasha-scale:


tests/helpers.py
```diff
@@ -120,6 +120,7 @@ def reference_id_from_url(url):
     for i in range(len(TEST_POLYGON_ANNOTATIONS))
 ]
 
+TEST_INDEX_EMBEDDINGS_FILE = "https://scale-ml.s3.amazonaws.com/tmp/sasha/pytest_embeddings_payload.json"
```
rkaplan, line 123: If this is stored in /tmp I believe it will get automatically deleted pretty soon, at which point the tests will fail. Unit test artifacts need to be in permanent storage.

tests/test_indexing.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_index_integration(dataset):
+    signed_embeddings_url = get_signed_url(TEST_INDEX_EMBEDDINGS_FILE)
+    create_response = dataset.create_custom_index(signed_embeddings_url)
+    assert JOB_ID_KEY in create_response
+    assert MESSAGE_KEY in create_response
+    job_id = create_response[JOB_ID_KEY]
+
+    # Job can error because pytest dataset fixture gets deleted
```
rkaplan, line 48: Why does this cause the job to error — wouldn’t the create_index request still go out successfully?

rkaplan (approved):
LGTM barring comments

PR #44, opened 2021-03-26
fix artifacts

sasha-scale:
Learned the S3 location I've been using to store testing artifacts is temporary - moved entities to new (permanent) file location


jihan-yin (approved):
so that's where the mscoco masks went 🤡

PR #48, opened 2021-04-06
change test files to S3 urls that are accessible

sasha-scale:
TL;DR: pytest is currently broken due to the same S3 permissioning challenges we've been seeing with ingestion.  The easiest fix I found was just changing the seed images to images from pandaset.

Previously, image uploads from urls were not being copied. Now that the scaleapi PR copies them, we run into a problem where we get access denied errors on BDD images. I switched to using Pandaset images for pytest (which are stored in a different S3 bucket) and it fixed the problem.

### DIFF
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -102,7 +102,7 @@ class Dataset:
         self,
         annotations: List[Union[BoxAnnotation, PolygonAnnotation]],
         update: Optional[bool] = DEFAULT_ANNOTATION_UPDATE_MODE,
-        batch_size: int = 20,
+        batch_size: int = 5000,
     ) -> dict:
         """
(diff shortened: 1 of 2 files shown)


nucleus/dataset.py
```diff
@@ -102,7 +102,7 @@ def annotate(
         self,
         annotations: List[Union[BoxAnnotation, PolygonAnnotation]],
         update: Optional[bool] = DEFAULT_ANNOTATION_UPDATE_MODE,
-        batch_size: int = 20,
+        batch_size: int = 5000,
```
rkaplan, line 105: Intentional?

sasha-scale: intentional, the batch_size is way too small.

ardila (approved):
These changes seem pretty safe to me

PR #49, opened 2021-04-08
batch uploads: raise Error

Nastia39:
* raise Error if `batch_size` argument is `< 0`
* pass `batch_size` for local upload as well

### DIFF
--- a/nucleus/annotation.py
+++ b/nucleus/annotation.py
@@ -90,7 +90,7 @@ class SegmentationAnnotation(Annotation):
     @classmethod
     def from_json(cls, payload: dict):
         return cls(
-            mask_url=payload.get(MASK_URL_KEY),
+            mask_url=str(payload.get(MASK_URL_KEY)),
             annotations=[
                 Segment.from_json(ann)
                 for ann in payload.get(ANNOTATIONS_KEY, [])
--- a/nucleus/model_run.py
+++ b/nucleus/model_run.py
@@ -135,7 +135,7 @@ class ModelRun:
             ):
                 if type_key in annotation_payload:
                     annotation_response[type_key] = [
-                        type_cls.from_json(ann)
+                        type_cls.from_json(ann)  # type: ignore
                         for ann in annotation_payload[type_key]
                     ]
             return annotation_response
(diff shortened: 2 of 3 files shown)


nucleus/annotation.py
```diff
@@ -90,7 +90,7 @@ def __str__(self):
     @classmethod
     def from_json(cls, payload: dict):
         return cls(
-            mask_url=payload.get(MASK_URL_KEY),
+            mask_url=str(payload.get(MASK_URL_KEY)),
```
ardila, line 93: Thanks for fixing this typing mistake!

I think we should raise an error if payload[MASK_URL_KEY] is None, instead of converting it to string and passing it on to cause problems elsewhere.

if MASK_URL_KEY not in payload:
  raise ValueError(f"{MASK_URL_KEY} field is required but was missing in the following annotation: {payload}")

ardila: then you can just do payload[MASK_URL_KEY] and the typing should be ok

nucleus/model_run.py
```diff
@@ -135,7 +135,7 @@ def _format_prediction_response(
             ):
                 if type_key in annotation_payload:
                     annotation_response[type_key] = [
-                        type_cls.from_json(ann)
+                        type_cls.from_json(ann)  # type: ignore
```
ardila, line 138: I had a more extensive fix that you might as well use


```
def _format_prediction_response(
        self, response: dict
    ) -> Union[
        dict,
        List[Union[BoxPrediction, PolygonPrediction, SegmentationPrediction]],
    ]:
        annotation_payload = response.get(ANNOTATIONS_KEY, None)
        SegmentationPrediction.from_json({"asdf": "asdf"})
        if not annotation_payload:
            # An error occurred
            return response
        annotation_response = {}
        type_key_to_class: Dict[
            str,
            Union[
                Type[BoxPrediction],
                Type[PolygonPrediction],
                Type[SegmentationPrediction],
            ],
        ] = {
            BOX_TYPE: BoxPrediction,
            POLYGON_TYPE: PolygonPrediction,
            SEGMENTATION_TYPE: SegmentationPrediction,
        }
        for type_key in annotation_payload:
            type_class = type_key_to_class[type_key]
            annotation_response[type_key] = [
                type_class.from_json(annotation)
                for annotation in annotation_payload[type_key]
            ]
        return annotation_response
```

ardila (changes requested):

nucleus/annotation.py
ardila: I address this in #53 so we could just merge that first?

nucleus/model_run.py
ardila: I address this in #53 so we could just merge that first?

PR #50, opened 2021-04-14
Switch to poetry to make publishing easy

ardila:
In order to publish now

poetry build

First publish to testpypi and check that things look ok at 
poetry config repositories.testpypi https://test.pypi.org/legacy/
poetry publish -r testpypi

Username: NucleusTeam
Password: Look it up in 1password (look for pypi)

If things look ok then

poetry publish

Things I am wondering: what tests were already passing, what is the best way to test this?
Is there a PR template to follow?


ardila:
Wait something went wrong with which branch I used for this, fixing so that not so many files are included

ardila:
Fixed in different PR

PR #53, opened 2021-04-14
Improve reprs for all objects in API

ardila:
Adds (or changes) all reprs so that

eval(str(test_object)) == test_object

I didn't feel sure about changing what equality meant for Model yet, but everything else I made an equality check that is as extensive as possible.

I added lots of tests for this and tested using

poetry run pytest. Some tests fail due to non-uniqueness issue with model ids in the database, but that happens in master too.

I tried to make things that feel like they are just data containers into dataclasses.

There was some pylint and typing stuff I fixed along the way as well.

Once we have #51 merged, that will make this PR smaller. If you want it split up further let me know.


ardila:
We should merge some of the other changes first (especially the poetry one) then I can clean things up a bit since I made this change on top of those (by accident)

nucleus/__init__.py
```diff
@@ -961,6 +975,42 @@ def delete_slice(self, slice_id: str) -> dict:
         )
         return response
 
+    def append_to_slice(
```
ardila, line 978: I moved this here to stop pylint complaining about directly accessing the client's private method

conftest.py
```diff
@@ -6,6 +6,7 @@
 # https://github.com/gevent/gevent/issues/1016#issuecomment-328530533
 # https://github.com/spyoungtech/grequests/issues/8
 import grequests
+
```
ardila, line 9: These are just changes due to black. If we want to remove them from this PR let me know

nucleus/model_run.py
```diff
@@ -125,19 +134,31 @@ def loc(self, dataset_item_id: str):
...
+                Type[BoxPrediction],
+                Type[PolygonPrediction],
+                Type[SegmentationPrediction],
+            ],
+        ] = {
+            BOX_TYPE: BoxPrediction,
+            POLYGON_TYPE: PolygonPrediction,
+            SEGMENTATION_TYPE: SegmentationPrediction,
+        }
+        for type_key in annotation_payload:
```
ardila, line 158: I added this so we stop getting mypy errors about this.

nucleus/prediction.py
```diff
@@ -91,9 +93,6 @@ def from_json(cls, payload: dict):
             metadata=payload.get(METADATA_KEY, {}),
         )
 
-    def __str__(self):
```
ardila, line 94: These are no longer needed because of inheritance from the base class

nucleus/slice.py
```diff
@@ -47,18 +51,9 @@ def append(
...
-
-        payload: Dict[str, Any] = {}
-        if dataset_item_ids:
-            payload[DATASET_ITEM_IDS_KEY] = dataset_item_ids
-        if reference_ids:
-            payload[REFERENCE_IDS_KEY] = reference_ids
-
-        response = self._client._make_request(
-            payload, f"slice/{self.slice_id}/append"
+        response = self._client.append_to_slice(
```
ardila, line 54: Moved the logic into the client to avoid direct access.

ardila:
Ok this should be ready for review!

nucleus/annotation.py
```diff
@@ -33,16 +41,11 @@ def from_json(cls, payload: dict):
             return SegmentationAnnotation.from_json(payload)
 
 
+@dataclass
```
sasha-scale, line 44: oh this is cool, never seen dataclass before.

Great for brevity!

nucleus/annotation.py
```diff
@@ -62,35 +65,25 @@ def to_payload(self) -> dict:
...
+    mask_url: str
+    annotations: List[Segment]
+    reference_id: Optional[str] = None
+    item_id: Optional[str] = None
+    annotation_id: Optional[str] = None
+
+    def __post_init__(self):
+        if not self.mask_url:
             raise Exception("You must specify a mask_url.")
-        if bool(reference_id) == bool(item_id):
```
sasha-scale, line 77: Is this deletion purposeful? I think UX hypothesis behind this check is that it's much better user experience to know your SegmentationAnnotation is malformed before sending the API request.

nucleus/prediction.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
sasha-scale, line 1: This import seems unused to me - can you elaborate on what it's needed for?

sasha-scale (approved):
LGTM, this is a big improvement for debugging!

nucleus/prediction.py
ardila: This allows for the type annotations in this file to work properly.

sasha-scale: why would it be needed here but not other files (like `nucleus/annotation.py`)

PR #54, opened 2021-04-15
Add version in global place

fbcotter:
Version didn't match between `setup.py` and `__init__.py`


fbcotter:
If this gets in, it would be excellent if you could tag and release versions

PR #59, opened 2021-04-21
higher timeout based on timeout settings on server-side

sasha-scale:
Playing around with the API endpoints, I was able to observe a timeout happen naturally, and was able to cause timeout errors when having the endpoint sleep before doing any computations.

The original purpose of introducing the timeout was to prevent connections from hanging indefinitely. The client side timeout should definitely be higher that the corresponding server side configuration, since it should never be the case that the client errors out when the server would have successfully serviced the request.

I think this should at least alleviate the user pain, although I would also want to understand more deeply why their upload was taking > 60 sec per batch


ardila:
LGTM.

ardila (approved):
The test errors are due to our server returning 500 on dataset deletion.

PR #61, opened 2021-04-23
Simplify error handling + Export Autotags

ardila:
* I decided to simplify our error handling in order to auto-surface errors from the backend. Currently we were burying or ignoring most. Eventually we can make the error classes more specific, but IMO anytime there is a bad response we should raise an error and print out what the backend thinks the issue is.
* I added tests showing that this surfaces the backend error, and that we get back autotag scores as expecated from the backend.
* The test can only run if you have access to the API_KEY for the default user used in our CircleCI setup, otherwise it will skip it.

As of Friday 1:12, this depends on a change that has not deployed yet and so it is expected to fail (I tested against a locally running backend server).

### DIFF
--- a/nucleus/__init__.py
+++ b/nucleus/__init__.py
@@ -167,7 +167,7 @@ class NucleusClient:
         Lists available models in your repo.
         :return: model_ids
         """
-        model_objects = self._make_request({}, "models/", requests.get)
+        model_objects = self.make_request({}, "models/", requests.get)
 
         return [
             Model(
@@ -185,14 +185,14 @@ class NucleusClient:
         Lists available datasets in your repo.
         :return: { datasets_ids }
         """
-        return self._make_request({}, "dataset/", requests.get)
+        return self.make_request({}, "dataset/", requests.get)
 
     def get_dataset_items(self, dataset_id) -> List[DatasetItem]:
         """
         Gets all the dataset items inside your repo as a json blob.
         :return [ DatasetItem ]
         """
-        response = self._make_request(
+        response = self.make_request(
             {}, f"dataset/{dataset_id}/datasetItems", requests.get
         )
         dataset_items = response.get("dataset_items", None)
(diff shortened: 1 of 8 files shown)


nucleus/__init__.py
```diff
@@ -1144,11 +1148,14 @@ def _make_request(
...
+            self.handle_bad_response(endpoint, requests_command, response)
 
-        if getattr(response, "status_code") not in SUCCESS_STATUS_CODES:
-            logger.warning(response)
+        return response.json()
 
-        return (
-            response.json()
-        )  # TODO: this line fails if response has code == 404
+    def handle_bad_response(self, endpoint, requests_command, response):
```
sasha-scale, line 1160: nice, much cleaner

nucleus/errors.py
```diff
@@ -22,3 +22,12 @@ class DatasetItemRetrievalError(Exception):
     def __init__(self, message="Could not retrieve dataset items"):
         self.message = message
         super().__init__(self.message)
+
+
+class NucleusAPIError(Exception):
```
sasha-scale, line 27: In the future, I think it would be really productive to overhaul error handling on the server side to return a sentry error_id here (similar to 500 errors in the UI) such that when a user reports a bug we can lookup the stack trace

tests/test_dataset.py
```diff
@@ -18,6 +25,8 @@
     DATASET_ID_KEY,
 )
 
+TEST_AUTOTAG_DATASET = "ds_bz43jm2jwm70060b3890"
```
sasha-scale, line 28: maybe include a note detailing which user this dataset belongs to

sasha-scale (approved):
LGTM barring nits

tests/test_dataset.py
ardila: Not sure I want to make that information public?

nucleus/errors.py
ardila: That would be nice!

nucleus/__init__.py
ardila: Yeah no more throwing informative error messages on the ground

PR #71, opened 2021-05-28
Da/async annotations

ardila:
* Python client changes to support asynchronous annotations
* Update package version to prepare for pushing
* Integration tests (including a common error case: wrong/missing reference ids)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
           name: Pytest Test Cases
           command: | # Run test suite, uses NUCLEUS_TEST_API_KEY env variable
             mkdir test_results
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           name: Pytest Test Cases
           command: | # Run test suite, uses NUCLEUS_TEST_API_KEY env variable
             mkdir test_results
-            poetry run coverage run --include=nucleus/* -m pytest --junitxml=test_results/junit.xml
+            poetry run coverage run --include=nucleus/* -m pytest -s -v --junitxml=test_results/junit.xml
```
ardila, line 38: Made the tests more verbose so they don't timeout as easily when step functions is slow about scheduling.

sasha-scale (approved):
LGTM

PR #73, opened 2021-06-04
Add prediction class pdf support

jihan-yin:
Adds support for uploading the entire class probability distribution for each prediction object with updated tests.


ardila (approved):
Can you fix the tests before merging? Should just be a matter of updating the tests actually. There was a backend change where I stopped sending out the errors in the response and instead log them to the job errors table, so just have to delete that field from the expected response.

PR #74, opened 2021-06-11
Remove Grequests Dependency

ardila:
* Remove grequests as a dependency
* Switch to using aiohttp for local file uploads
* Add warning for large uploads
* Local file uploads will now crash loudly if anything goes wrong on any batch instead of silently proceeding with errors
* Disallow sending NaNs for now since that would mess up querying
* Bump version to prepare for release
* Match all dependency requirements with colab so we stop getting errors when trying to use the client in colab

The default behavior is now for all synchronous API endpoints (local and remote) to crash on any failure, and for asynchronous large batch uploads to proceed but log errors.

### DIFF
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -225,7 +225,7 @@ class Dataset:
         return self._client.populate_dataset(
             self.id,
             dataset_items,
-            force=update,
+            update=update,
             batch_size=batch_size,
         )
 
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -216,7 +216,10 @@ class Dataset:
 
         if len(dataset_items) > WARN_FOR_LARGE_UPLOAD and not asynchronous:
             print(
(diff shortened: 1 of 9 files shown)


nucleus/dataset.py
```diff
@@ -211,6 +214,11 @@ def append(
         """
         check_for_duplicate_reference_ids(dataset_items)
 
+        if len(dataset_items) > WARN_FOR_LARGE_UPLOAD and not asynchronous:
+            print(
+                "WARNING: for large uploads we recommend uploading your images remotely to a cloud storage provider and then using asynchronous=True."
```
rkaplan, line 219: I would rewrite this as:

"Tip: for large uploads, get faster performance by importing your images into Nucleus directly from a cloud storage provider. See https://dashboard.scale.com/nucleus/docs/api?language=python#guide-for-large-ingestions for details."

rkaplan (approved):
LGTM barring changing the warning copy

nucleus/dataset.py
ardila: Thanks!

PR #75, opened 2021-06-16
Nucleus - prediction pdf optional

jihan-yin:
Makes the prediction pdf truly optional.


ardila (approved):
Pretty sure this is correct. Can you bump the version in pyproject.toml so we can release this?

PR #76, opened 2021-06-17
Batch export functionality for slice and dataset + point fix

ardila:
* Add support for batch export to slice and dataset
* Add missing point class (dict still supported) and better local errors

### DIFF
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -4,6 +4,7 @@ import requests
 
 from nucleus.job import AsyncJob
 from nucleus.utils import (
+    convert_export_payload,
     format_dataset_item_response,
     serialize_and_write_to_presigned_url,
 )
@@ -327,3 +328,23 @@ class Dataset:
 
     def check_index_status(self, job_id: str):
         return self._client.check_index_status(job_id)
+
+    def items_and_annotations(
+        self,
+    ) -> List[Dict[str, Union[DatasetItem, Dict[str, List[Annotation]]]]]:
+        """Returns a list of all DatasetItems and Annotations in this slice.
+
+        Returns:
+            A list, where each item is a dict with two keys representing a row
+            in the dataset.
+            * One value in the dict is the DatasetItem, containing a reference to the
+                item that was annotated.
+            * The other value is a dictionary containing all the annotations for this
+                dataset item, sorted by annotation type.
+        """
+        api_payload = self._client.make_request(
+            payload=None,
+            route=f"dataset/{self.id}/exportForTraining",
+            requests_command=requests.get,
+        )
+        return convert_export_payload(api_payload["exportedRows"])
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -17,6 +17,7 @@ from .constants import (
     DATASET_NAME_KEY,
     DATASET_SLICES_KEY,
     DEFAULT_ANNOTATION_UPDATE_MODE,
+    EXPORTED_ROWS,
     JOB_ID_KEY,
     NAME_KEY,
     REFERENCE_IDS_KEY,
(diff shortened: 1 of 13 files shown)


nucleus/slice.py
```diff
@@ -109,42 +110,12 @@ def items_and_annotations(
             * The other value is a dictionary containing all the annotations for this
                 dataset item, sorted by annotation type.
         """
-        return list(self.items_and_annotation_generator())
-
-    def annotate(
-        self,
```
ardila, line 115: this is useless, better to annotate from dataset

sasha-scale: agreed

tests/helpers.py
```diff
@@ -115,7 +117,7 @@ def reference_id_from_url(url):
     {
         **TEST_POLYGON_ANNOTATIONS[i],
         "confidence": 0.10 * i,
-        "class_pdf": TEST_POLYGON_MODEL_PDF,
```
sasha-scale, line 118: seems to me like we should keep this test pdf to ensure test coverage on the pdf upload feature.

nucleus/slice.py
```diff
@@ -109,42 +110,12 @@ def items_and_annotations(
...
-            dataset_id=self.dataset_id,
-            annotations=annotations,
-            update=update,
-            batch_size=batch_size,
+        api_payload = self._client.make_request(
+            payload=None,
+            route=f"slice/{self.slice_id}/exportForTraining",
+            requests_command=requests.get,
         )
+        return convert_export_payload(api_payload["exportedRows"])
```
sasha-scale, line 118: :nit: let's make exportedRows a constant

tests/helpers.py
```diff
@@ -149,10 +151,10 @@ def assert_polygon_annotation_matches_dict(
         annotation_instance.annotation_id == annotation_dict["annotation_id"]
     )
     for instance_pt, dict_pt in zip(
-        annotation_instance.vertices, annotation_dict["vertices"]
+        annotation_instance.vertices, annotation_dict["geometry"]["vertices"]
```
sasha-scale, line 154: :nit: user constants like VERTICES_KEY here

nucleus/dataset.py
```diff
@@ -327,3 +328,23 @@ def delete_custom_index(self):
...
+                item that was annotated.
+            * The other value is a dictionary containing all the annotations for this
+                dataset item, sorted by annotation type.
+        """
+        api_payload = self._client.make_request(
+            payload=None,
+            route=f"dataset/{self.id}/exportForTraining",
+            requests_command=requests.get,
+        )
+        return convert_export_payload(api_payload["exportedRows"])
```
sasha-scale, line 350: constant here too

sasha-scale (changes requested):
Code looks good to me, just a few small comments wrt to cleanup.  Testing locally, the test_slice.py::test_slice_append test fails for me.

Seems like the root of the issue is that the slice contains 3 items but `slc.items_and_annotations()` returns an empty array, which seems unexpected.  Can you take a second look at that?

tests/helpers.py
ardila: There was a bug I was testing here (when no class pdf, it would error out. BUt I think Jihan separately added a test case for that)

tests/helpers.py
ardila: That goes against style everywhere else in the test, will just keep consistent for now

nucleus/dataset.py
ardila: Done

nucleus/slice.py
ardila: Done

sasha-scale (approved):
LGTM

PR #81, opened 2021-06-29
Batch export for model predictions

ardila:
* Add support for batch exporting a model run

### DIFF
--- a/tests/test_prediction.py
+++ b/tests/test_prediction.py
@@ -50,7 +50,7 @@ def test_reprs():
 def model_run(CLIENT):
     ds = CLIENT.create_dataset(TEST_DATASET_NAME)
     ds_items = []
-    for url in TEST_IMG_URLS:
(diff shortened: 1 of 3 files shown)


tests/test_prediction.py
```diff
@@ -50,7 +50,7 @@ def test_repr(test_object: any):
 def model_run(CLIENT):
     ds = CLIENT.create_dataset(TEST_DATASET_NAME)
     ds_items = []
-    for url in TEST_IMG_URLS:
+    for url in TEST_IMG_URLS[:2]:
```
sasha-scale, line 53: remove slicing?

sasha-scale (approved):
LGTM barring nit

tests/test_prediction.py
ardila: It makes the tests faster and they all still pass :)

sasha-scale: haha ok fine w/ me

PR #83, opened 2021-07-02
add cuboid annotations to client

drew-kaul:
This PR adds client side support for CuboidAnnotations and CuboidPredictions. The endpoint for uploading predictions and annotations already supports cuboids, so uploading cuboid annotations and predictions to a dataset from the Python client should work now.

Edit: Backend changes for /exportForTraining have landed

Changed DATASET_ITEM_ID_KEY to ITEM_ID_KEY in annotation/prediction classes since backend expects the field "item_id" to be populated

### DIFF
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -12,7 +12,6 @@ from nucleus.utils import (
 
 from .annotation import (
     Annotation,
-    CuboidAnnotation,
     check_all_mask_paths_remote,
 )
 from .constants import (
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -12,7 +12,6 @@ from nucleus.utils import (
 
 from .annotation import (
     Annotation,
-    CuboidAnnotation,
     check_all_annotation_paths_remote,
 )
 from .constants import (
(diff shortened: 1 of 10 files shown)


nucleus/dataset.py
```diff
@@ -168,9 +167,6 @@ def annotate(
             "ignored_items": int,
         }
         """
-        if any((isinstance(ann, CuboidAnnotation) for ann in annotations)):
```
ardila, line 171: still need this check in dataset.annotate since you can't upload them yet

ardila (approved):
LGTM but even I'm not clear on what is supported or not.

Can you at least clarify in the PR description what is expected to work and what isn't?

nucleus/dataset.py
drew-kaul: cuboids can be uploaded, the backend changes are already there. here's the tests for that:
[tests](https://github.com/scaleapi/scaleapi/blob/master/test/nucleus/testEndpoints.ts#L786)

ardila: should add test coverage for this here then to make sure the format generated by the client is compatible with the backend

PR #85, opened 2021-07-15
Nucleus - custom indexing improvements

jihan-yin:
Changes python api for custom indexing to support new custom indexing api.


nucleus/__init__.py
```diff
@@ -1101,9 +1102,14 @@ def delete_model(self, model_id: str) -> dict:
         )
         return response
 
-    def create_custom_index(self, dataset_id: str, embeddings_url: str):
+    def create_custom_index(
```
song-william, line 1105: Consider adding a doc string here. Methods defined here are part of the public API.

tests/test_indexing.py
```diff
@@ -39,7 +39,7 @@ def dataset(CLIENT):
 
 def test_index_integration(dataset):
     signed_embeddings_url = TEST_INDEX_EMBEDDINGS_FILE
-    create_response = dataset.create_custom_index(signed_embeddings_url)
+    create_response = dataset.create_custom_index([signed_embeddings_url])
```
song-william, line 42: Is this call missing the `embedding_dimension` argument?

song-william (approved):
Approving now to unblock while Jihan looks into the external dependencies that are breaking these test. Also, it might be bad practice to put binary blobs into the git history, as it could stick around in the tree forever.

jihan-yin:
Tests pass locally and circleci seems flaky, merging for now

PR #86, opened 2021-07-22
delete annotations

sasha-scale:
Add support for annotation deletion via Python client


nucleus/__init__.py
```diff
@@ -1037,6 +1038,28 @@ def delete_slice(self, slice_id: str) -> dict:
         )
         return response
 
+    def delete_annotations(
+        self, dataset_id: str, reference_ids: list = None, keep_history=False
```
ardila, line 1042: Thanks for doing this... I said I would do it and then I got distracted by other things. You da best

ardila (approved):
Tests look great, Feel free to merge once this goes green

tests/test_annotation.py
```diff
@@ -302,3 +302,19 @@ def test_polygon_gt_upload_ignore(dataset):
     assert_polygon_annotation_matches_dict(
         response_annotation, TEST_POLYGON_ANNOTATIONS[0]
     )
+
+    @pytest.mark.integration
```
ardila, line 306: way to catch this detail. helps a lot for local development to be able to skip these tests that spin up step functions

PR #91, opened 2021-08-02
Resolve cuboid yaw from geometry dict

flubstep:
The yaw parameter was previously being read incorrectly from the annotation payload, instead of being pulled from the result of the annotation's geometry. This caused all the annotations and predictions to assume yaw of 0. This PR resolves the key access.


ardila:
Merge master and rerun tests please, I think that should fix everything

ardila (approved):

PR #93, opened 2021-08-02
Fix local upload jankiness

ardila:
* Fix nested asyncio bug by using nest-asyncio
* Actually pass in the update field for local upload which we weren't doing before


song-william (approved):
Tests should become green after this PR merges: https://github.com/scaleapi/scaleapi/pull/28274

PR #96, opened 2021-08-06
Implement Scene class

drew-kaul:
Implement Scene class and related classes (Frame, SceneDatasetItem, LidarScene)
Perform validation and construct payload
Refactor integration tests for sync upload
Add async tests

Tests will pass after this lands: https://github.com/scaleapi/scaleapi/pull/28716 

Refactor SceneDatasetItem to DatasetItem


nucleus/scene.py
```diff
@@ -0,0 +1,78 @@
...
+    frames: List[Frame]
+    reference_id: str
+    metadata: Optional[dict] = None
+
+
+@dataclass
+class LidarScene(Scene):
+    def __post_init__(self):
+        # do validation here for lidar scene
+        lidar_sources = flatten(
```
sasha-scale, line 64: I think a user would actually construct the scene over time i.e. initialize a scene object, call scene.add_item() repeatedly (or something roughly along those lines) and then upload via API.  I therefore think validation is something that should happen on upload rather than post-init.

sasha-scale (commented):
good start, added some preliminary commentary.

In order to test end-to-end you can overload `dataset.append` in the following way: 

if the list submitted is typeof(LidarScene), submit to upload_scenes URL, otherwise submit to append URL

nucleus/scene.py
```diff
@@ -0,0 +1,78 @@
...
+from typing import Optional, Dict, List, Set
+from enum import Enum
+from .annotation import Point3D
+from .utils import flatten
+
+
+class DatasetItemType(Enum):
+    IMAGE = "image"
+    POINTCLOUD = "pointcloud"
+    VIDEO = "video"
```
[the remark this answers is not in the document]
drew-kaul: Not sure what you mean by this - to validate payload, we need to check type is one of these strings. Are you saying I should remove `VIDEO` from the enum?

sasha-scale (commented):
You should also add support for hitting the upload_scenes endpoint.  Using diego's API design, that would go something like:

in __init__.py file:
```
def append_scenes(self, dataset_id, scenes): 
```

and then add some logic in dataset.append to determine whether to call `populate_dataset` vs `append_scenes`.  i.e. instead of `dataset_items: List[DatasetItem]` the input would be `items: List[Union[DatasetItem, LidarScene]]`

nucleus/scene.py
```diff
@@ -0,0 +1,147 @@
...
+
+@dataclass
+class LidarScene(Scene):
+    def validate(self):
+        lidar_sources = flatten(
+            [
+                [
+                    source
+                    for source in frame.items.keys()
+                    if frame.items[source].type == DatasetItemType.POINTCLOUD
```
sasha-scale, line 129: this is not interpreted as a string comparison for some reason
```suggestion
                   item for frame.items.values() if str(item.type) == DatasetItemType.POINTCLOUD
```

nucleus/scene.py
```diff
@@ -49,18 +78,31 @@ def __post_init__(self):
                 value, SceneDatasetItem
             ), "All values must be SceneDatasetItems"
 
+    def add_item(self, item: SceneDatasetItem, sensor_name: str):
+        self.items[sensor_name] = item
+
 
 @dataclass
 class Scene:
```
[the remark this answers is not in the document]
drew-kaul: made Scene abstract

nucleus/scene.py
drew-kaul: actually this check will fail - tested it in my terminal. Since the right hand side is of type `DatasetItemType`, I have to compare `frame.items[source].type` rather than `str()` of that

drew-kaul: both sides have to be of type `DatasetItemType`, so I casted the left side to `DatasetItemType`

drew-kaul: changed `type` param to string and perform validation to check one of enum values - resolves issue

nucleus/scene.py
drew-kaul: moved to separate function - will call before upload

nucleus/scene.py
```diff
@@ -0,0 +1,274 @@
...
+            assert isinstance(
+                value, SceneDatasetItem
+            ), "All values must be SceneDatasetItems"
+
+    def add_item(self, item: SceneDatasetItem, sensor_name: str):
+        self.items[sensor_name] = item
+
+    def to_payload(self) -> dict:
+        return {
+            INDEX_KEY: self.index,
```
sasha-scale, line 150: see comment on API side - remove index field here

drew-kaul: resolved

nucleus/dataset.py
```diff
@@ -248,6 +263,34 @@ def append(
             batch_size=batch_size,
         )
 
+    def append_scenes(
+        self,
+        scenes: List[LidarScene],
+        update: Optional[bool] = False,
+        asynchronous: Optional[bool] = False,
+    ) -> Union[dict, AsyncJob]:
+        """TODO: Add updated docstring here"""
```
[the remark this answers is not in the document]
sasha-scale: similar to the append function you should probably add some warning here to use the asynchronous endpoint if you have more than 5 scenes, or something along those lines

sasha-scale (changes requested):
A couple minor comments - looks good! My opinion is that we should fold SceneDatasetItem into the DatasetItem class, and make type default to "image" and camera_params optional.  Curious to see what Diego thinks on this point

nucleus/dataset.py
```diff
@@ -248,6 +263,34 @@ def append(
...
+            return AsyncJob.from_json(response, self._client)
+
+        payload = construct_append_scenes_payload(scenes, update)
+        response = self._client.make_request(
+            payload=payload,
+            route=f"{self.id}/upload_scenes",
+        )
+        return response
+
     def upload_scenes(
```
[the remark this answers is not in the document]
drew-kaul: already changed existing unit tests, just forgot to get rid of `upload_scenes` - fixed

nucleus/scene.py
```diff
@@ -0,0 +1,299 @@
...
+            payload[CAMERA_PARAMS_KEY] = self.camera_params.to_payload()
+        return payload
+
+    def to_json(self) -> str:
+        return json.dumps(self.to_payload(), allow_nan=False)
+
+
+@dataclass
+class Frame:
+    index: Union[int, None] = None
```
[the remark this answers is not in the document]
drew-kaul: fixed

nucleus/dataset.py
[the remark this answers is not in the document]
drew-kaul: added docstring and warning for async

drew-kaul:
> A couple minor comments - looks good! My opinion is that we should fold SceneDatasetItem into the DatasetItem class, and make type default to "image" and camera_params optional. Curious to see what Diego thinks on this point

drew-kaul:
> Implement Scene class and related classes (Frame, SceneDatasetItem, LidarScene)
> Perform validation and construct payload
> Refactor integration tests for sync upload
> Add async tests
> 
> Tests will pass after this lands: [scaleapi/scaleapi#28716](https://github.com/scaleapi/scaleapi/pull/28716)
> 
> Refactor SceneDatasetItem to DatasetItem

All scene upload related tests pass

nucleus/scene.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def flatten(t):
+    return [item for sublist in t for item in sublist]
+
+
+def check_all_scene_paths_remote(scenes: List[LidarScene]):
+    for scene in scenes:
+        for frame in scene.frames_dict.values():
+            for item in frame.items.values():
```
sasha-scale, line 176: This only works for pointclouds and will error out on images.  Here's the fix:

```
import IMAGE_LOCATION_KEY from constants.py

def check_all_scene_paths_remote(scenes: List[LidarScene]):
    for scene in scenes:
        for frame in scene.frames_dict.values():
            for item in frame.items.values():
                pointcloud_location = getattr(item, POINTCLOUD_LOCATION_KEY)
                if pointcloud_location and is_local_path(pointcloud_location):
                    raise ValueError(
                        f"All paths for DatasetItems in a Scene must be remote, but {item.pointcloud_location} is either "
                        "local, or a remote URL type that is not supported."
                    )
                image_location = getattr(item, IMAGE_LOCATION_KEY)
                if image_location and is_local_path(image_location):
                    raise ValueError(
                        f"All paths for DatasetItems in a Scene must be remote, but {item.image_location} is either "
                        "local, or a remote URL type that is not supported."
                    )
```

tests/helpers.py
```diff
@@ -33,16 +33,16 @@
             "frames": [
                 {
                     "lidar": {
-                        "url": TEST_POINTCLOUD_URLS[0],
-                        "type": "pointcloud",
+                        "pointcloud_url": TEST_POINTCLOUD_URLS[0],
                         "reference_id": "lidar_frame_1",
                         "metadata": {},
                     },
```
sasha-scale, line 39: Can we make this frame a bit more complex - i.e. one lidar pointcloud and one image with camera_params? I think it's important to test the image upload functionality in the integration test as well

nucleus/dataset_item.py
```diff
@@ -40,28 +139,39 @@ def from_json(cls, payload: dict):
...
+                self.image_location
+            ), "Must specify image_location for DatasetItems not in a Scene"
+            payload[IMAGE_URL_KEY] = self.image_location
+
         if self.reference_id:
             payload[REFERENCE_ID_KEY] = self.reference_id
         if self.item_id:
             payload[DATASET_ITEM_ID_KEY] = self.item_id
+        if self.camera_params:
+            payload[CAMERA_PARAMS_KEY] = self.camera_params
```
sasha-scale, line 164:
```suggestion
            payload[CAMERA_PARAMS_KEY] = self.camera_params.to_payload()
```

sasha-scale (approved):
LGTM barring my two outstanding comments.  Ship it!

nucleus/dataset.py
```diff
@@ -199,16 +200,16 @@ def ingest_tasks(self, task_ids: dict):
 
     def append(
```
ardila, line 201: What will happen if you try to append dataset items to a 3D dataset? I thought we were going to ban this

nucleus/dataset_item.py
```diff
@@ -2,36 +2,135 @@
...
+class DatasetItem:  # pylint: disable=R0902
+    image_location: Optional[str] = None
     reference_id: Optional[str] = None
     item_id: Optional[str] = None
     metadata: Optional[dict] = None
+    pointcloud_location: Optional[str] = None
 
     def __post_init__(self):
         self.local = is_local_path(self.image_location)
+        assert bool(self.image_location) != bool(
```
ardila, line 100: Need to check that exactly one of 

image_location
pointcloud_location

is passed

nucleus/dataset_item.py
```diff
@@ -2,36 +2,135 @@
...
+            else None
+        )
+        self.camera_params = (
+            CameraParams.from_json(camera_params) if camera_params else None
+        )
 
     @classmethod
-    def from_json(cls, payload: dict):
-        url = payload.get(IMAGE_URL_KEY, "") or payload.get(
+    def from_json(cls, payload: dict, is_scene=False):
```
ardila, line 118: This object should be serializable/deserializable independent of caller location.

Have to figure out a way to make this work without passing a parameter.

I think you can always do the is_scene logic but have to change the pointcloud url and image url to both default to none.
cls(
image_url = payload.get(IMAGE_URL_KEY, None) or payload.get(ORIGINAL_IMAGE_URL, None)
pointcloud_location = payload.get(POINTCLOUD_URL_KEY, None)....

nucleus/dataset_item.py
```diff
@@ -40,28 +139,39 @@ def from_json(cls, payload: dict):
...
-        payload = {
-            IMAGE_URL_KEY: self.image_location,
+    def to_payload(self, is_scene=False) -> dict:
+        payload: Dict[str, Any] = {
             METADATA_KEY: self.metadata or {},
         }
+
+        if is_scene:
+            if self.image_location:
+                payload[URL_KEY] = self.image_location
```
ardila, line 149: Same comment as above: change the logic so serialization works properly regardless of whether this is a scene or not.

nucleus/dataset_item.py
drew-kaul: I believe that this code does this - ` bool(self.image_location) != bool(self.pointcloud_location)` essentially performs an XOR so only 1 has been set

nucleus/dataset.py
sasha-scale: this requires changes on the scaleapi side, we currently do not track the notion of 3d vs. 2d dataset in postgres.  I'd recommend considering this out of scope for this particular PR

nucleus/dataset_item.py
drew-kaul: The serialization will work properly regardless of whether it's a scene, but the payload that's returned will be different. I believe the default to None is handled bc in the `__post_init__` we enforce that only one of image_url, pointcloud_url can be valid and the other is None

nucleus/dataset.py
drew-kaul: will need to add dataset validation for this in follow up PR

nucleus/dataset_item.py
drew-kaul: same as above - serialization will work in both cases, but the payloads produced are different (since the `append` and `upload_scenes` endpoints expect diff formats

nucleus/dataset_item.py
ardila: The on-the-wire format should not depend on whether it is a scene. It's possible to have just one piece of logic that works for both, and it would be much cleaner.

nucleus/dataset_item.py
ardila: Same as above: have only one JSON format for this class not two

PR #97, opened 2021-08-07
remove need to pass in dataset id

ardila:
It wasn't needed anywhere and simplifies API + makes the documentation up to date again.


nucleus/utils.py
```diff
@@ -144,18 +144,18 @@ def upload_to_presigned_url(presigned_url: str, file_pointer: IO):
...
-    dataset_id: str,
+    request_id: str,
     client,
 ):
-    request_id = uuid.uuid4().hex
+    request_uuid = uuid.uuid4().hex
     response = client.make_request(
         payload={},
-        route=f"dataset/{dataset_id}/signedUrl/{request_id}",
+        route=f"dataset/{request_id}/signedUrl/{request_uuid}",
```
jihan-yin, line 153: If this path is going to be able to accept ids that aren't just dataset ids, we need to change the routing. Maybe something like `presignRequest/request_id` or something.

jihan-yin (approved):
LGTM, let's add a todo + ticket to track changing the endpoint path

nucleus/utils.py
ardila: I don't want to break old clients though hmmm

PR #98, opened 2021-08-11
API for continuous indexing and manual image indexing

jihan-yin:


ardila (changes requested):
We need an integration test for this on a small dataset

ardila:
The test is failing, but once it's green this looks ok

ardila (approved):

PR #101, opened 2021-08-14
Add getters to scene class

drew-kaul:
Add getters to access parts of a LidarScene (frame, sensors) or Frame (items)
Add unit tests for all methods


sasha-scale:
Unit tests would be a great addition, I also recommend adding `__repr__` methods to to the classes (which makes it more meaningful to print an instance of a scene/frame class

drew-kaul:
> Unit tests would be a great addition, I also recommend adding `__repr__` methods to to the classes (which makes it more meaningful to print an instance of a scene/frame class

Added unit tests for all Frame/Scene methods. I don't think adding a `__repr__` method is necessary - DatasetItem doesn't have one, and by printing out a Scene, you get all the fields there anyways (or you could you use the getters)

sasha-scale (approved):
LGTM

PR #104, opened 2021-08-19
Small tweaks to improve Frame UX

sasha-scale:
Took Diego's suggestion in stride and changed the Frame constructor.  In the process I simplified the frame abstraction by removing the index field.  If a user initializes a Scene with a list of frames, I think it's reasonable to infer frame position from list position.

### DIFF
--- a/nucleus/scene.py
+++ b/nucleus/scene.py
@@ -53,7 +53,7 @@ class Frame:
             sensor: DatasetItem.from_json(item, is_scene=True)
             for sensor, item in payload.items()
         }
-        return cls(items=items)
+        return cls(**items)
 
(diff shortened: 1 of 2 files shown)


ardila (approved):

nucleus/scene.py
```diff
@@ -15,10 +15,11 @@
 from .dataset_item import DatasetItemType, DatasetItem
 
 
-@dataclass
 class Frame:
```
ardila, line 18: Don't skip making this a dataclass, just add keyword args besides items to the items dict

ardila: The downside to this approach is you lose __repr__, eq and many other convenience functions

PR #109, opened 2021-08-27
Trivial changes to make privacy mode work

ardila:


nucleus/constants.py
```diff
@@ -86,6 +86,7 @@
 TYPE_KEY = "type"
 UPDATED_ITEMS = "updated_items"
 UPDATE_KEY = "update"
+UPLOAD_TO_SCALE_KEY = "uploadToScale"
```
sasha-scale, line 89: should be snake_case, will be converted to camelCase on typescript side

tests/test_dataset.py
```diff
@@ -176,8 +178,11 @@ def check_is_expected_response(response):
 
     # Plain image upload
     ds_items_plain = []
-    for url in TEST_IMG_URLS:
-        ds_items_plain.append(DatasetItem(image_location=url))
+    for i, url in enumerate(TEST_IMG_URLS):
+        upload_to_scale = i == 0
```
sasha-scale, line 182: useful to leave an explanatory comment here, i.e. (upload first item in privacy mode)

sasha-scale (changes requested):
Looks like there's a small bug - running `poetry run pytest tests/test_dataset.py::test_append_and_export` against the scaleapi feature branch fails :/

sasha-scale (approved):

PR #110, opened 2021-08-27
No more null reference ids

ardila:
Disallow python client from sending invalid requests: we no longer use dataset item id in append, and require reference id.


sasha-scale (approved):
LGTM

PR #112, opened 2021-09-02
Add Autocurate to API, and add test

fractal1729:
This PR adds functionality for users to run acquisition functions on their model predictions via `nucleus.autocurate.entropy`. This triggers the `nucleus_autocurate_acquisition` step function. There is an integration test in `tests/test_autocurate.ts` which requires the sfn and backend servers to be running locally.

There is also a minor change in promoting the `modelRun.dataset_id` attribute from `modelRun._dataset_id` so it can be accessed in other places.

### DIFF
--- a/nucleus/autocurate.py
+++ b/nucleus/autocurate.py
@@ -19,6 +19,7 @@ def entropy(name, model_runs, client):
         route=f"autocurate/{dataset_id}/single_model_entropy/{name}",
         requests_command=requests.post,
     )
+    # TODO: the response should already have the below three fields populated
     response[JOB_LAST_KNOWN_STATUS_KEY] = "Started"
     response[JOB_TYPE_KEY] = "autocurateEntropy"
     response[JOB_CREATION_TIME_KEY] = (
--- a/nucleus/autocurate.py
+++ b/nucleus/autocurate.py
@@ -1,0 +1,25 @@
+import datetime
+import requests
+from nucleus.constants import (
+    JOB_CREATION_TIME_KEY,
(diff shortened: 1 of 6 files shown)


nucleus/autocurate.py
```diff
@@ -0,0 +1,25 @@
...
+    JOB_CREATION_TIME_KEY,
+    JOB_LAST_KNOWN_STATUS_KEY,
+    JOB_TYPE_KEY,
+)
+from nucleus.job import AsyncJob
+
+
+def entropy(name, model_runs, client):
+    model_run_ids = [model_run.model_run_id for model_run in model_runs]
+    dataset_id = model_runs[0].dataset_id
```
ardila, line 13: assert len({_.dataset_id for _ in model_runs}) == 0, f"Model runs have conflicting dataset ids: {model_runs}"

nucleus/autocurate.py
```diff
@@ -0,0 +1,25 @@
...
+    response = client.make_request(
+        payload={"modelRunIds": model_run_ids},
+        route=f"autocurate/{dataset_id}/single_model_entropy/{name}",
+        requests_command=requests.post,
+    )
+    response[JOB_LAST_KNOWN_STATUS_KEY] = "Started"
+    response[JOB_TYPE_KEY] = "autocurateEntropy"
+    response[JOB_CREATION_TIME_KEY] = (
```
ardila, line 21: return job object from backend

ardila (approved):
good to check in once we push daily deployu

PR #120, opened 2021-09-13
Load testing + id cleanups

ardila:
While adding a script for load testing, I found a lot of annoyance in 2 things so I fixed them

1. We should auto-populate null annotation ids to make inserts idempotent without requiring users to pass in annotation ids
2. We need to get rid of dataset item id everywhere in the client


nucleus/__init__.py
```diff
@@ -1111,8 +1110,7 @@ def delete_annotations(
     def append_to_slice(
         self,
         slice_id: str,
-        dataset_item_ids: List[str] = None,
-        reference_ids: List[str] = None,
+        reference_ids: List[str],
```
song-william, line 1113: nit: I think you need to update the param doc string to no longer include `dataset_item_ids`

nucleus/dataset.py
```diff
@@ -353,8 +352,7 @@ def loc(self, dataset_item_id: str) -> dict:
     def create_slice(
         self,
         name: str,
-        dataset_item_ids: List[str] = None,
-        reference_ids: List[str] = None,
+        reference_ids: List[str],
```
song-william, line 355: nit: same here about updating doc string

nucleus/slice.py
```diff
@@ -147,7 +145,7 @@ def export_embeddings(
 
 def check_annotations_are_in_slice(
     annotations: List[Annotation], slice_to_check: Slice
-) -> Tuple[bool, Set[str], Set[str]]:
+) -> Tuple[bool, Set[str]]:
```
song-william, line 148: nit: update return value docs

scripts/load_test.py
```diff
@@ -0,0 +1,199 @@
+from nucleus import annotation
```
song-william, line 1: Should we consider moving this into our private repo?

song-william (approved):
Mainly doc string nits. Do you use an IDE extension that can automatically help check this?

scripts/load_test.py
ardila: I think it's fine as reference

ardila:
> Mainly doc string nits. Do you use an IDE extension that can automatically help check this?

Do you know of one? I don't

PR #121, opened 2021-09-16
Support for new autotag exporting apis

jihan-yin:
https://github.com/scaleapi/scaleapi/pull/30319


nucleus/constants.py
```diff
@@ -62,7 +63,8 @@
 MODEL_RUN_ID_KEY = "model_run_id"
 NAME_KEY = "name"
 NEW_ITEMS = "new_items"
-NUCLEUS_ENDPOINT = "https://api.scale.com/v1/nucleus"
+NUCLEUS_ENDPOINT = "https://jihan-yin-api.devbox.internal.scale.com/v1/nucleus"
```
ardila, line 66: You can set this using an environment variable by the way

I have this in my bash_profile

```
alias api_local="export NUCLEUS_ENDPOINT=\"http://localhost:3000/v1/nucleus\" && echo $NUCLEUS_ENDPOINT"
alias api_prod="export NUCLEUS_ENDPOINT=\"https://api.scale.com/v1/nucleus\" && echo $NUCLEUS_ENDPOINT"
```

nucleus/dataset.py
```diff
@@ -86,21 +87,36 @@ def items(self) -> List[DatasetItem]:
...
+    def autotag_items(self, autotag_name, for_scores_greater_than=0):
+        """For a given Autotag of this dataset, export its tagged items with scores above a threshold, largest scores first.
 
-        If you have pandas installed, you can create a pandas dataframe using
+        :return: dictionary of the form
+            {
+                'autotagItems': AutotagItem[],
+                'autotag': Autotag
+            }
+        See https://dashboard.nucleus.scale.com/nucleus/docs/api#export-autotag-items for more details on the return types.
```
ardila, line 98: Document return type here so that you can do

help(dataset.autotag_items) and directly see the return format

ardila: (I know this is annoying and we should auto-generate API docs from these docstrings, we are working towards that and not there yet though, sorry)

nucleus/dataset.py
```diff
@@ -86,21 +87,36 @@ def items(self) -> List[DatasetItem]:
...
-        pandas.Dataframe(dataset.autotag_scores(autotag_name))
+    def autotag_training_items(self, autotag_name):
+        """For a given Autotag of this dataset, export its training items. These are user selected positives during refinement.
 
         :return: dictionary of the form
-            {'ref_ids': List[str],
-             'datset_item_ids': List[str],
-             'score': List[float]}
+            {
+                'autotagTrainingItems': AutotagTrainingItem[],
```
ardila, line 112: Either define this class in the client or document this

ardila (changes requested):
Needs an integration test, you can see the existing one for export autotag for reference. We have a test dataset with autotag already set up.

nucleus/constants.py
jihan-yin: sick didn't see that we grab it from the environmental variables now

ardila (approved):

PR #129, opened 2021-09-27
Update custom indexing integration test to sleep until the job is complete

jihan-yin:
Otherwise, the dataset gets completed before the job is finished and results in errors being reported on our side.


ardila:
This test is very flaky right now. I actually made this change already and had to revert it.

So I guess right now we have 3 options

1. Skip it
2. Get noise in indexing logs
3. Get noise in integration test logs + occasionally block PRs in this repo


IMO #2 is the best option until we have time to look into why this test is flaky OR have the step function properly handle that error case without creating noise

PR #130, opened 2021-09-29
Get Scene API

ardila:
Added test for new endpoint
and some eq methods to make that test easier

### DIFF
--- a/nucleus/dataset_item.py
+++ b/nucleus/dataset_item.py
@@ -122,24 +122,16 @@ class DatasetItem:  # pylint: disable=R0902
         )
 
     @classmethod
-    def from_json(cls, payload: dict, is_scene=False):
+    def from_json(cls, payload: dict):
         image_url = payload.get(IMAGE_URL_KEY, None) or payload.get(
             ORIGINAL_IMAGE_URL_KEY, None
         )
-
-        if is_scene:
-            return cls(
-                image_location=image_url,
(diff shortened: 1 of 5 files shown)


nucleus/dataset_item.py
```diff
@@ -122,24 +122,16 @@ def __post_init__(self):
         )
 
     @classmethod
-    def from_json(cls, payload: dict, is_scene=False):
+    def from_json(cls, payload: dict):
```
song-william, line 125: Is removing this an argument an API breaking change?

tests/test_scene.py
```diff
@@ -260,6 +260,10 @@ def test_scene_upload_sync(dataset):
 
     response = dataset.append(scenes, update=update)
 
+    first_scene = dataset.get_scene(scenes[0].reference_id)
+
+    assert first_scene == scenes[0]
+
```
song-william, line 266: Consider adding a simple case to ensure that `!=` also works.

song-william (approved):

sasha-scale (approved):

nucleus/dataset_item.py
ardila: This is almost entirely used internally so I don't think so

tests/test_scene.py
ardila: Done

PR #131, opened 2021-10-06
Model-only endpoints

ardila:
This should be enough in terms of endpoints to no longer have to use the model run concept.

I want to get this PR approved, push out the client, then notify all users before rewriting all of our docs just because it's such a big change.

### DIFF
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -533,7 +533,7 @@ class Dataset:
         )
 
     def export_predictions(self, model):
-        """Exports all predications from a model on this dataset"""
+        """Exports all predictions from a model on this dataset"""
         json_response = self._client.make_request(
             payload=None,
             route=f"dataset/{self.id}/model/{model.id}/export",


nucleus/dataset.py
```diff
@@ -525,3 +531,134 @@ def get_scene(self, reference_id) -> Scene:
...
+            "allowed_label_matches": List[AllowedMatch],
+        }"""
+        if options is None:
+            options = {}
+        return self._client.make_request(
+            payload=options,
+            route=f"dataset/{self.id}/model/{model.id}/calculateEvaluationMetrics",
+        )
+
+    def upload_predictions(
```
sasha-scale, line 572: I think there should be parity in naming between predictions and annotations.  If this func is named `upload_predictions` maybe `annotate` should be `upload_annotations`?  Or vice versa, `predict`, `annotate`

ardila: I wanted to differentiate between the eventual

model.predict

and 

dataset.upload_predictions

To be clear that actual prediction is happening in one case and in another prediction is actually running on hosted model inference.

PR #133, opened 2021-10-12
Fix all python tests

ardila:


conftest.py
```diff
@@ -30,3 +30,13 @@ def dataset(CLIENT):
     yield ds
 
     CLIENT.delete_dataset(ds.id)
+
```
ardila, line 33: Added this code in case we accidentally delete the test dataset again it will be easier to fix.

PR #149, opened 2021-11-09
[Model CI] Full client support for current functionality

phil-scale:
Adds Python client support for the following
- Creating unit tests and unit test metrics
- Evaluating a model on unit tests
- Getting the metrics associated with a unit test
- Getting the evaluation history of a unit test

This PR also introduces a `modelci` folder and a `ModelCIClient` that inherits from the base `NucleusClient`. This is an effort to make the relationship between Nucleus and Model CI directed, where Model CI currently depends on Nucleus but Nucleus does not need to depend on Model CI. The separate folder and client should also allow for easier refactoring in the future.

[[sc-262746]](https://app.shortcut.com/scaleai/story/262746)
[[sc-269906]](https://app.shortcut.com/scaleai/story/269906)


nucleus/modelci/utils.py
```diff
@@ -0,0 +1,42 @@
+from .unit_test_evaluation import (
+    UnitTestEvaluation,
+    UnitTestItemEvaluation,
+    UnitTestEvaluationStatus,
+)
+
+
+def try_convert_float(float_str: str) -> float:
```
yixu34, line 8: `NaN` has a [special meaning in the definition of floats](https://en.wikipedia.org/wiki/NaN), e.g. when you do `0 / 0`. I think we want to surface some other kind of error string if someone passes in something that wasn't even a valid float to begin with.

tests/modelci/test_unit_test.py
```diff
@@ -0,0 +1,104 @@
...
+
+    unit_test_metric = MODELCI_CLIENT.create_unit_test_metric(
+        unit_test_name=test_name,
+        eval_function_name=EVAL_FUNCTION_NAME,
+        threshold=EVAL_FUNCTION_THRESHOLD,
+        threshold_comparison=ThresholdComparison.GREATER_THAN,
+    )
+    assert unit_test_metric["unit_test_id"]
+    assert unit_test_metric["eval_function_id"]
+    assert unit_test_metric["threshold"]
```
yixu34, line 64: Might want to be explicit here because python treats zeroes as false-y.

tests/modelci/test_unit_test.py
```diff
@@ -0,0 +1,104 @@
...
+    assert unit_test_metric["threshold_comparison"]
```
yixu34, line 65: Some of the other fields might be auto-generated, or it'd be inappropriate to peek into implementation details, but in this case, it should be OK to check for the `GREATER_THAN` that you passed in, right?

tests/modelci/test_unit_test.py
```diff
@@ -0,0 +1,104 @@
...
+    assert metrics[0] == unit_test_metric
+
+
+# TODO: Move unit test to fixture once deletion is implemented
+def test_unit_test_metric_creation_from_class(MODELCI_CLIENT, dataset):
+    # create some dataset_items for the unit test to reference
+    items = make_dataset_items()
+    dataset.append(items)
+    test_name = (
+        "unit_test_" + str(uuid.uuid4())[-N_UUID_CHARACTERS:]
```
yixu34, line 79: Wonder how the Nucleus folks feel about creating a util for this `str(uuid.uuid4())[-N_UUID_CHARACTERS:]` pattern that keeps coming up.

nucleus/modelci/unit_test_evaluation.py
```diff
@@ -0,0 +1,40 @@
...
+from enum import Enum
+from dataclasses import dataclass
+from typing import List
+
+
+@dataclass
+class UnitTestEvaluationStatus(str, Enum):
+    """Job status of unit test evaluation."""
+
+    PENDING = ("pending",)
```
yixu34, line 11: Curious why we need to write this as a single-item tuple 🤔 

Seems neat that we're treating it as both `str` and `Enum` though.

yixu34 (commented):
A concrete class inheriting from another is generally a smell. Here's one article that talk about it: https://www.rangakrish.com/index.php/2016/08/07/is-deriving-from-a-concrete-class-bad/

One way to think of it is through the Liskov substitution principle. Seems slightly weird IMO to say that "a ModelCI client is-a Nucleus client". It also means that we're now surfacing two clients in the client library. Like I get the idea at the moment, but things might diverge in unexpected ways down the line.

But, seeing as how generally we're invested in ModelCI leveraging Nucleus, I would almost think of ModelCI as a mixin/expansion pack to the Nucleus client library. This leverages composition instead of inheritance. So it'd look something like:

```python3
class ModelCIClientMixin:
    <functionality>
```
https://docs.github.com/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax
```python3
class NucleusClient(ModelCIClientMixin):
```

Or if you want to do it more dynamically/explicitly, you could pass it through a constructor:
```python3
class NucleusClient:
    def __init__(self, addons=...)
```

nucleus/modelci/unit_test_evaluation.py
phil-scale: Not sure, this is copied from code that @sasha-scale previously wrote.

nucleus/modelci/utils.py
phil-scale: I do think we should still fail gracefully, but maybe I can just log an error as well?

yixu34: Not sure if we need to log stuff in user code, that should be on them (it does bring up the question of what kind of API we want to provide for user-defined logging, if any). Though if we're provided any pre-built code, then I think it can make sense to log ourselves.

We should still be able to fail gracefully while still being type-correct. I think this stems from `UnitTestItemEvaluation` having `result` being a `float`, when IMO it should be an `Optional[float]`. The db table should correspondingly have the column be `nullable`, to signify an exception in the code (maybe you'd need an extra `nullable` column for an error code/string). Overloading `NaN` to mean both an exception and `0 / 0` could be a confusing edge case.

phil-scale: Ah ok `Optional[float]` makes sense - sounds good.

tests/modelci/test_unit_test.py
phil-scale: Yup thanks!

tests/modelci/test_unit_test.py
phil-scale: Yup, that makes sense, thanks!

tests/modelci/test_unit_test.py
phil-scale: Added a helper method for this.k

nucleus/modelci/unit_test_evaluation.py
phil-scale: I just changed them to strings, which I think should be ok.

yixu34:
Ok to recap our offline discussion today around my [earlier comment on inheritance and whatnot](https://github.com/scaleapi/nucleus-python-client/pull/149#pullrequestreview-802104064):
* Mixin inheritance looks weird. So instead, we can use proper composition (i.e. `NucleusClient` takes in a `ModelCIClient` in the `NucleusClient` constructor).
* Alternatively: the main reason why we tried inheritance in the first place was to reuse the low-level request code. That can be refactored into a "neutral" module that is used by both `NucleusClient` and `ModelCIClient`. Since we're just using IDs between the two systems (i.e. they don't use the same client dataclasses), it's safe to do this.
* However, the Sia team wrote up some thoughts on the design of the Nucleus client library, wherein they propose that we move more in the direction of proper client data classes, which might make the above point trickier: https://scaleapi.slack.com/archives/G019J8773HA/p1635766488004600

For now, we can go with the current class design, but we should start a discussion asap.

nucleus/modelci/__init__.py
```diff
@@ -0,0 +1,257 @@
...
+        )
+        return [UnitTestMetric(**metric) for metric in response["metrics"]]
+
+    def evaluate_model_on_unit_tests(
+        self, model_id: str, unit_test_names: List[str]
+    ) -> AsyncJob:
+        """Evaluates the given model on the specified Unit Tests. ::
+
+            import nucleus.modelci as nm
+            client = nm.NucleusClient("YOUR_SCALE_API_KEY")
```
sasha-scale, line 189:
```suggestion
            client = nm.ModelCIClient("YOUR_SCALE_API_KEY")
```

sasha-scale (changes requested):
Can you add pytest coverage for client.list_unit_tests?

Before checking this in, take a look at the autogenerated docs by running:

```
poetry update
cd docs
poetry run sphinx-autobuild . ./_build/html --watch ../nucleus
```

There's some funkiness going on - most notably that due to the class inheritance, all nucleus client documentation is dunked into the model_ci page.  There's probably some sphinx setting to diable this behavior that we should take a look at, I recommend reaching out to Drake.

In general with code snippets in the documentation, it's important to check that they run successfully before shipping - we've gotten into trouble with example code that doesn't actually run in the past (its an easy mistake to make 😓)

phil-scale:
Thanks for all the helpful comments! The tests are failing because the evaluation function `IOU` is inaccessible to the test account, so I think I'll first implement the fix for eval functions before continuing with this PR.

phil-scale:
Hey @sasha-scale and @yixu34 - I went with making `ModelCI` an "extension" to the original `NucleusClient`. I'm not sure that this is the right approach to go with, but I think it's fairly flexible and easy to change at this point. Would love to get your feedback of this interface, using `nucleus_client.model_ci.create_unit_test` and so on.

nucleus/connection.py
```diff
@@ -0,0 +1,71 @@
...
+        :param payload: given payload
+        :param route: route for the request
+        :param requests_command: requests.post, requests.get, requests.delete
+        :return: response JSON
+        """
+        endpoint = f"{self.endpoint}/{route}"
+
+        logger.info("Make request to %s", endpoint)
+
+        for retry_wait_time in RetryStrategy.sleep_times:
```
yixu34, line 42: I would probably just import some retry library to help with this. I see that we use `tenacity` in `models` FWIW. Doing an exponential backoff with base 3 instead of base 2 is a bit quirky 😛

yixu34 (commented):
FWIW I think `nucleus_client.modelci.create_unit_test` is OK as an interface, especially if we're thinking of Nucleus as the overall platform that users interact with. Curious what @sasha-scale thinks.

Also, we touched upon this during sprint planning, but this PR is getting a bit "horizontal" in the sense that we're lumping a bit too many changes in here that coincidentally happen to touch the client library. I'd say let's freeze the scope right now and try and push this through. My main ask is to consider a library for retries; will def to @sasha-scale for the rest.

nucleus/connection.py
phil-scale: I'll add a comment for this as a TODO. I don't want to lump that into this PR - this code was directly copied from the Nucleus base python client, so I think that we could introduce a change there in an individual PR.

phil-scale: Created this ticket to track: https://app.shortcut.com/scaleai/story/281630/use-external-library-for-http-retry-logic

nucleus/__init__.py
```diff
@@ -42,7 +42,7 @@
     ERROR_ITEMS,
     ERROR_PAYLOAD,
     ERRORS_KEY,
-    EVAL_FUNCTION_NAME_KEY,
+    ID_KEY,
```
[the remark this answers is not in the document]
phil-scale: Fixed, thanks!

nucleus/constants.py
```diff
@@ -41,13 +41,13 @@
...
 ERROR_ITEMS = "upload_errors"
 ERROR_PAYLOAD = "error_payload"
-EVAL_FUNCTION_NAME_KEY = "eval_function_name"
 FRAMES_KEY = "frames"
 FX_KEY = "fx"
 FY_KEY = "fy"
 GEOMETRY_KEY = "geometry"
 HEADING_KEY = "heading"
 HEIGHT_KEY = "height"
+ID_KEY = "id"
```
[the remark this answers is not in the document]
phil-scale: Yup, removed - thanks!

nucleus/modelci/unit_test.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        return self._client.create_unit_test_metric(
+            self.name, eval_function_name, threshold, threshold_comparison
+        )
+
+    def get_metrics(self) -> List[UnitTestMetric]:
+        """Retrieves all metrics of the Unit Test. ::
+
+            import nucleus.modelci as nm
+            client = nm.NucleusClient("YOUR_SCALE_API_KEY")
```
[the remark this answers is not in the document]
phil-scale: Updated to just `nucleus.NucleusClient`, and we can use `nucleus.modelci.create_unit_test`

nucleus/modelci/__init__.py
```diff
@@ -0,0 +1,257 @@
...
+
+        Returns:
+            A list of UnitTest objects.
+        """
+        response = self.make_request(
+            {},
+            "modelci/unit_test/list",
+            requests_command=requests.get,
+        )
+        return [UnitTest(test_id) for test_id in response["unit_test_ids"]]
```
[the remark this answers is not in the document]
phil-scale: Fixed, thanks!

nucleus/modelci/__init__.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def get_unit_test_metrics(self, unit_test_id: str) -> List[UnitTestMetric]:
+        """Retrieves all metrics of the Unit Test. ::
+
+            import nucleus.modelci as nm
+            client = nm.NucleusClient("YOUR_SCALE_API_KEY")
```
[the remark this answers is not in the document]
phil-scale: Same here

nucleus/modelci/__init__.py
phil-scale: Ditto

phil-scale:
Final pytest is failing because the endpoint for evaluation takes too long - I don't think that's a Python client problem, so I'll leave this test file as a separate PR: https://github.com/scaleapi/nucleus-python-client/pull/154

nucleus/modelci/__init__.py
```diff
@@ -0,0 +1,295 @@
...
+            {},
+            f"modelci/unit_test/{unit_test_id}",
+            requests_command=requests.delete,
+        )
+        return response["success"]
+
+    def evaluate_model_on_unit_tests(
+        self, model_id: str, unit_test_names: List[str]
+    ) -> AsyncJob:
+        """Evaluates the given model on the specified Unit Tests. ::
```
sasha-scale, line 224: Documentation is slightly incorrect here, if the unit_test_names are omitted, the model will be evaluated on all unit tests.  The array of names is a subsampling method, if that makes sense

sasha-scale (approved):
Some suggestions for improved usability:
- Improve string representations of classes to be consistent with other classes, here's an example:
https://github.com/scaleapi/nucleus-python-client/blob/2eb3aaf9fc1401febba682980dbcc14196713d86/nucleus/dataset.py#L69

Looks like `unit_test.get_eval_history()` doesn't work, should fix before merge

nucleus/modelci/__init__.py
phil-scale: Hmm I think that no matter what, the model will only be evaluated on the specified unit tests, since `unit_test_names` is not optional.

In the case that it's the empty list, we would pass an empty list into the `TestEvaluator` constructor, and it looks to me like the subsampling is only checking if the input list is null, not if it's empty (see https://github.com/scaleapi/scaleapi/blob/d994d223ca552aee814f7c8fadbe26a6e6ee49bb/server/models/nucleus/modelci/MCUnitTest.ts#L209-L211)

phil-scale: Opened up `node` just to confirm
```
> tests = []
[]
> if (tests) console.log("hello")
hello
```

nucleus/modelci/__init__.py
```diff
@@ -0,0 +1,295 @@
...
+            "modelci/eval_fn",
+            requests_command=requests.get,
+        )
+        return EvalFunction(**response["eval_functions"])
+
+    def create_unit_test(self, name: str, slice_id: str) -> UnitTest:
+        """Creates a new Unit Test. ::
+
+            import nucleus
+            from nucleus.modelci.unit_test import ThresholdComparison
```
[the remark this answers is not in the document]
phil-scale: Thanks!

nucleus/connection.py
```diff
@@ -0,0 +1,71 @@
...
+logging.basicConfig()
+
+
+# TODO: use retry library instead of custom code. Tenacity is one option.
+class RetryStrategy:
+    statuses = {503, 504}
+    sleep_times = [1, 3, 9]
+
+
+class Connection:
```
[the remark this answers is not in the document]
phil-scale: Yeah credit to @yixu34 for this!

phil-scale:
I just tried locally and `get_eval_history` works with me - it's `evaluate_model_on_unit_tests` that is problematic for me, but I think the problem is on the `scaleapi` side

```
>>> unit_tests = client.modelci.list_unit_tests()
>>> unit_tests[0].get_eval_history()
[]
>>> print(unit_tests[0])
UnitTest(unit_test_id='ut_c6ay39xmav9g048azy30', client=ModelCI(connection='Connection(api_key='scaleint_57d7743357384b9184e1030d16751880|615cc6e424ecf90020baffdd', endpoint='https://api.scale.com/v1/nucleus')')
>>>
```

tests/test_slice.py
```diff
@@ -173,7 +172,7 @@ def test_slice_send_to_labeling(dataset):
 
     # Slice creation
     slc = dataset.create_slice(
-        name=(TEST_SLICE_NAME + str(uuid.uuid4())[-N_UUID_CHARACTERS:]),
+        name=(TEST_SLICE_NAME + get_uuid()),
```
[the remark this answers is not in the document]
phil-scale: <3

nucleus/modelci/__init__.py
```diff
@@ -0,0 +1,295 @@
...
+
+        Returns:
+            Whether deletion was successful.
+        """
+        response = self._connection.make_request(
+            {},
+            f"modelci/unit_test/{unit_test_id}",
+            requests_command=requests.delete,
+        )
+        return response["success"]
```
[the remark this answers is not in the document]
phil-scale: done, thanks!

nucleus/modelci/__init__.py
```diff
@@ -0,0 +1,295 @@
...
+            A list of UnitTestEvaluation objects
+        """
+        response = self._connection.make_request(
+            {},
+            f"modelci/unit_test/{unit_test_id}/eval_history",
+            requests_command=requests.get,
+        )
+        return [
+            UnitTestEvaluation(**eval[ID_KEY])
+            for eval in response["evaluations"]
```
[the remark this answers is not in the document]
phil-scale: done, thanks!

PR #152, opened 2021-11-12
Make dataset properties more responsive

gatli:
- Query dataset/details for list of datasets
- Introduce Pydantic and use that for list_datasets
- Add tests for replaced dataset property access


gatli:
Included in https://github.com/scaleapi/nucleus-python-client/pull/153

PR #157, opened 2021-11-27
Fix merge error -> recursion bug in dataset.info

gatli:
A merge error slipped into last PR. `dataset.info` was not moved properly. 

This PR fixes that the request is handled in the `Dataset` class.


jean-lucas (approved):
LGTM! Quick fix.

Do you need to update a patch version of the package?

gatli:
> LGTM! Quick fix.
> 
> Do you need to update a patch version of the package?

Yes! Good catch 🙂

PR #158, opened 2021-11-30
[Model CI] Update Model CI python API to be cleaner and more object-oriented

phil-scale:
Update Model CI Python API according to these principles: https://docs.google.com/document/d/1Z0PFgvxz3zW5xNn97WdRCq7VeqP5vzyljv95Zv8GpVg/

Removed the following methods from the `ModelCI` client add-on:
* `create_unit_test_metric`
* `get_unit_test_info`
* `get_unit_test_eval_history`

Updated and fixed docs.

PR #159, opened 2021-11-30
Create unit test metrics with unit tests

gatli:
This PR introduces creating unit test metrics alongside the unit test. It also cleans up the ModelCI client according to our object oriented principles (sorry, should probably have made another PR 😢 ) which Phil documented earlier.

A part of this was creating a more "Pythonic" way of defining Metrics for a test.

I've introduced the class `AvailableEvalFunctions` as a collection object for `EvalFunctions`. This collection object acts as an entry to discover available standard (and in the future custom) evaluation functions through the SDK. 

Standard Scale provided evaluation functions are created as an attribute on this class and should be documented in the SDK allowing for nice IDE documentation. This allows us to define evaluation conditions (newly introduced term, plz challenge if you don't like it) for unit tests in a pythonic manner via statements like `iou() > 0.5` which then translates to the data transfer object `EvalFunctionCondition(id="ef_...", threshold_comparison=ThresholdComparison.greater_than, threshold=0.5)` which hides away the details of the REST API.

In the future this also creates a simple interface to add more parameters to evaluation functions via something like `iou(classes=["pedestrian", "cyclist"]) > 0.5`.

The new way of adding metrics to unit tests is via the `create_unit_test` method (notice the `evaluation_metrics`)
```
e = client.modelci.eval_functions
client.modelci.create_unit_test(
        name=test_name,
        slice_id=slc.slice_id,
        evaluation_metrics=[e.iou() > 0.5],
    )
```
or via `UnitTest.add_metric`
```
e = client.modelci.eval_functions
iou = e.iou
unit_test: UnitTest = ...
unit_test.add_metric(iou() > 0.5)
```

Some of the dataclasses have been translated into `pydantic.BaseModel` classes that allow us to parse the REST structures into Pythonic structures that take care of validation as well. This avoids having to create specialized `format_<class>` structures for nested dictionaries and allows us a clear place for data validation over the wire.


nucleus/modelci/__init__.py
```diff
@@ -83,81 +105,16 @@ def create_unit_test(self, name: str, slice_id: str) -> UnitTest:
...
+                name=name,
+                slice_id=slice_id,
+                evaluation_conditions=evalulation_conditions,
+            ).dict(),
             "modelci/unit_test",
-            requests_command=requests.post,
         )
         return UnitTest(response[UNIT_TEST_ID_KEY], self)
 
-    def create_unit_test_metric(
```
gatli, line 96: Moved to `UnitTest`

nucleus/modelci/__init__.py
```diff
@@ -83,81 +105,16 @@ def create_unit_test(self, name: str, slice_id: str) -> UnitTest:
...
-            requests_command=requests.post,
-        )
-        return UnitTestMetric(
-            unit_test_id=response[UNIT_TEST_ID_KEY],
-            eval_function_id=response[EVAL_FUNCTION_ID_KEY],
-            threshold=threshold,
-            threshold_comparison=threshold_comparison,
-        )
-
-    def get_unit_test_info(self, unit_test_id: str) -> UnitTestInfo:
```
gatli, line 145: Move to `UnitTest`

nucleus/modelci/__init__.py
```diff
@@ -172,40 +129,13 @@ def list_unit_tests(self) -> List[UnitTest]:
...
-            {},
+        response = self._connection.get(
             "modelci/unit_test",
-            requests_command=requests.get,
         )
         return [
             UnitTest(test_id, self) for test_id in response["unit_test_ids"]
         ]
 
-    def get_unit_test_metrics(self, unit_test_id: str) -> List[UnitTestMetric]:
```
gatli, line 184: Move to `UnitTest`

nucleus/modelci/__init__.py
```diff
@@ -252,54 +180,8 @@ def evaluate_model_on_unit_tests(
...
-            {},
-            f"modelci/unit_test/{unit_test_id}/eval_history",
-            requests_command=requests.get,
-        )
-        return [
-            self.get_unit_test_eval_info(eval[ID_KEY])
-            for eval in response[EVALUATIONS_KEY]
-        ]
-
-    def get_unit_test_eval_info(
```
gatli, line 289: Moved to `UnitTest`

nucleus/modelci/eval_function.py
```diff
@@ -1,14 +0,0 @@
-"""Data types for Evaluation Functions."""
```
gatli, line 1: Moved to `nucleus/modelci/data_transfer_objects/eval_function.py`

nucleus/modelci/unit_test.py
```diff
@@ -119,4 +117,43 @@ def get_eval_history(self) -> List[UnitTestEvaluation]:
...
+            f"modelci/unit_test/{self.id}/eval_history",
+        )
+        # TODO(gunnar): Repeated info calls are slow -> Move work to backend
+        eval_history = GetEvalHistory.parse_obj(response)
+        return [
+            self.get_unit_test_eval_info(evaluation.id)
+            for evaluation in eval_history.evaluations
+        ]
+
+    def info(self):
```
gatli, line 130: Moved from ModelCI `client`

nucleus/modelci/unit_test.py
```diff
@@ -119,4 +117,43 @@ def get_eval_history(self) -> List[UnitTestEvaluation]:
...
+
+        Returns:
+            A UnitTestInfo object
+        """
+        response = self._client._connection.get(
+            f"modelci/unit_test/{self.id}/info",
+        )
+        return UnitTestInfo(**response)
+
+    def get_unit_test_eval_info(
```
gatli, line 144: Moved from ModelCI `client`

nucleus/modelci/__init__.py
```diff
@@ -48,25 +51,46 @@ def __repr__(self):
...
 
-    def list_eval_functions(self) -> List[EvalFunction]:
+    @property
+    def eval_functions(self) -> AvailableEvalFunctions:
+        """List all available evaluation functions. ::
+
+        import nucleus
+        client = nucleus.NucleusClient("YOUR_SCALE_API_KEY")
+
+        eval_functions = client.modelci.eval_functions
```
phil-scale, line 61: nit: add tab to this - I think it needs an extra tab space to show up as code in the docs

tests/helpers.py
```diff
@@ -14,6 +14,7 @@
 TEST_DATASET_3D_NAME = "[PyTest] Test Dataset 3D"
 TEST_SLICE_NAME = "[PyTest] Test Slice"
 TEST_PROJECT_ID = "60b699d70f139e002dd31bfc"
+# What is this magic?
 TEST_EVAL_FUNCTION_ID = "ef_c61595wh49km7ppkk14g"
```
phil-scale, line 18: Yeah in fact we don't need this anymore! Now that `list_eval_functions` is good, you can just use the first eval function from there.

nucleus/modelci/eval_functions/available_eval_functions.py
```diff
@@ -0,0 +1,81 @@
...
+class MeanAveragePrecisionForBoxes(BaseEvalFunction):
+    """Mean average precision for boxes fitting the provided parameters"""
+
+    @classmethod
+    @property
+    def name(self) -> str:
+        return "mean_average_precision_for_boxes"
+
+
+EvalFunction = Union[Type[IoU], Type[MeanAveragePrecisionForBoxes]]
```
phil-scale, line 28: Why not just use `BaseEvalFunction`?

nucleus/modelci/eval_functions/available_eval_functions.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class AvailableEvalFunctions:
+    """Collection class that acts as a common entrypoint to access evaluation functions. Standard evaluation functions
+    provided by Scale are attributes of this class.
+
+    Examples:
+        e = client.modelci.eval_functions
+        unit_test_metrics = [e.iou() > 5, e.map() > 0.95, e.custom["customer_function"]() == True]
+    """
```
phil-scale, line 38: I get the idea here, but I think it's a bit confusing and unscalable to have a separate class that acts as a collection of all the eval functions. I'd prefer to keep a native list for consistency since all of our other "list" methods return native lists.

nucleus/modelci/data_transfer_objects/eval_function.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class EvalFunctionCondition(BaseModel):
+    """
+    eval_function_id: ID of evaluation function
+    threshold_comparison: comparator for evaluation. i.e. threshold=0.5 and threshold_comparator > implies that a test only passes if score > 0.5.
+    threshold: numerical threshold that together with threshold comparison, defines success criteria for test evaluation.
+    """
+
+    # TODO: Having only eval_function_id hurts readability -> Add function name
```
phil-scale, line 15: We initially went with eval function name but this quickly becomes a problem because of the concept of public eval functions. For each user, we enforce uniqueness of their evaluation function names, but we do not enforce global uniqueness of names, which means that public eval functions created by other users may have conflicting names.

We could address that by also enforcing uniqueness of names among a user's eval functions and public eval functions - I think that should be the way to go.

phil-scale (commented):
We're writing docstrings in a way that can look nice on the autogenerated documentation based on our comments. You can preview docs by running `poetry run sphinx-autobuild docs docs/_build/html --watch nucleus` and you can see your own docs at http://localhost:8000/api/nucleus/modelci/index.html

I think the eval functions interface is a bit confusing here because there are so many different variations of the concept of an eval function with `EvalFunctionEntry`, `BaseEvalFunction`, `AvailableEvalFunctions`, `GetEvalFunctions`, `EvalFunctionCondition`, and the union `EvalFunction`. I think we should remove a few and combine a few others if possible.

Finally, I happened to do a lot of the same refactoring here: https://github.com/scaleapi/nucleus-python-client/pull/158 but without any new functionality. I wonder if we can either merge that one into this one or merge that one into `model_ci_release` first and rebase this on top of that.

nucleus/modelci/eval_functions/available_eval_functions.py
phil-scale: I think in this case, the functionality to map by name is sort of optional for users, and we can provide optional functionality in the form of util functions.

phil-scale:
Also, wdyt of introducing faux immutability to pydantic classes via https://pydantic-docs.helpmanual.io/usage/models/#faux-immutability

nucleus/modelci/eval_functions/available_eval_functions.py
gatli: MyPy wasn't happy with the base class for some reason 🤷‍♂️

nucleus/modelci/eval_functions/available_eval_functions.py
gatli: A list or a dictionary doesn't play well at all with an IDE or Jupyter Notebook. You can't return a named entity without having an aggregator class. 

In this instance the `eval_functions` property behaves more like a namespaced module than a list ... and I think that's good and Pythonic.

When we add new evaluation functions we should add them as named objects here as well to make sure that the Jupyter notebook/IDE knows what available eval functions there are with normal autocomplete functionality. The abstraction `AvailableEvalFunctions` should barely be seen by the user, they only interact with its attributes through the modelci class.

I actually would like to change our other list methods into returning a query object as well, similar to this one. That allows for a lot more flexibility in interacting with the backend in a fast manner. 

I think this interface is very user-friendly and easier to write than more REST like APIs so I suggest we try this out and if we get bad feedback we can roll back to a simpler model.

nucleus/modelci/data_transfer_objects/eval_function.py
gatli: Yeah, the name problem seems to keep popping up. We need to think about how we can make it as nice as we can for the user. These IDs are simply non-readable so if you're stepping through your code you have no idea what this eval function is. I don't know what the best way to go about this but we'll figure it out 🙂

gatli:
> I think the eval functions interface is a bit confusing here because there are so many different variations of the concept of an eval function with `EvalFunctionEntry`, `BaseEvalFunction`, `AvailableEvalFunctions`, `GetEvalFunctions`, `EvalFunctionCondition`, and the union `EvalFunction`. I think we should remove a few and combine a few others if possible.

This is how you build abstraction hierarchies with  `pydantic`, this reflects how many different payload types we have communicating with the backend. Ideally we don't expose the `data_transfer_objects` to the user and rather have Model objects that they interact with. But I agree the clarity could be better. I think we just need to take another stab at the naming and granularity of these. Maybe they'll also get better when people are familiar with `pydantic`.

> Also, wdyt of introducing faux immutability to pydantic classes via https://pydantic-docs.helpmanual.io/usage/models/#faux-immutability

Very much in favour of that 🙂 


> Finally, I happened to do a lot of the same refactoring here: #158 but without any new functionality. I wonder if we can either merge that one into this one or merge that one into `model_ci_release` first and rebase this on top of that.

I merged it into `model_ci_release` and will rebase on top 🙌

gatli:
@phil-scale I have yet to go through the Sphinx docs and merge with your changes from the releaase branch. I'll finish that tomorrow.

Just to reiterate:
I really would like to try out the `AvailableEvalFunctions` approach as I think it is pretty neat. With bigger deployments I think query objects are also the future for other "list-like" functions as currently it is pretty hard to query for certain attributes of entities.

nucleus/modelci/unit_test.py
```diff
@@ -5,52 +5,21 @@
...
-    """
-
-    unit_test_id: str
-    eval_function_id: str
-    threshold: float
-    threshold_comparison: ThresholdComparison
+@dataclass
+class UnitTestInfo:
+    name: str
+    slice_id: str
```
phil-scale, line 22: I actually got rid of this class because the user never uses this, and it felt a little silly to just have a class of this form for no real purpose. I'm wondering if we can just get rid of the `info()` method because it's not super useful, and users will never call it either.

nucleus/modelci/unit_test.py
phil-scale: Yeah I think we can delete this method. It's currently unused, and users will never call it either.

tests/modelci/test_eval_function.py
```diff
@@ -1,6 +1,3 @@
-from nucleus.modelci.eval_function import EvalFunction
-
-
 def test_list_eval_functions(CLIENT):
-    eval_functions = CLIENT.modelci.list_eval_functions()
-    assert all(isinstance(eval_fn, EvalFunction) for eval_fn in eval_functions)
+    eval_functions = CLIENT.modelci.eval_functions
+    print(eval_functions.iou > 0.5)
```
phil-scale, line 3: Any way we could turn this into an assertion?

tests/modelci/test_unit_test.py
```diff
@@ -2,29 +2,24 @@
...
-    )
+    iou = CLIENT.modelci.eval_functions.iou
+    unit_test_metric = unit_test.add_criteria(iou() > EVAL_FUNCTION_THRESHOLD)
     assert unit_test_metric.unit_test_id == unit_test.id
     assert unit_test_metric.eval_function_id
     assert unit_test_metric.threshold == EVAL_FUNCTION_THRESHOLD
     assert unit_test_metric.threshold_comparison == EVAL_FUNCTION_COMPARISON
 
-    metrics = unit_test.get_metrics()
+    metrics = unit_test.get_criteria()
```
phil-scale, line 20: nit: can we also rename the variable to `criteria`?

phil-scale (commented):
Overall very nice, adding a few more comments here!

nucleus/modelci/__init__.py
```diff
@@ -19,33 +32,38 @@ class ModelCI:
...
-    def create_unit_test(self, name: str, slice_id: str) -> UnitTest:
+        payload = GetEvalFunctions.parse_obj(response)
+        return AvailableEvalFunctions(payload.eval_functions)
+
+    def create_unit_test(
+        self,
+        name: str,
+        slice_id: str,
+        evaluation_criteria: List[EvaluationCriteria],
+    ) -> UnitTest:
```
phil-scale, line 66: As suggested by @pfmark, maybe we should force the list of `evaluation_criteria` to be nonempty

nucleus/modelci/eval_functions/available_eval_functions.py
pfmark: I think it's a bit annoying that we need to add info about new eval functions at many different places (new class, add to `EvalFunctions`, add to `AvailableEvalFunctions`). But I guess whenever we try not to have the eval functions as member of `AvailableEvalFunctions`, IDEs won't like it, right?

nucleus/modelci/data_transfer_objects/eval_function.py
```diff
@@ -0,0 +1,41 @@
+from typing import List, Optional
+
+from pydantic import BaseModel, validator
+
+from ..constants import ThresholdComparison
+
+
+class EvaluationCriteria(BaseModel):
```
pfmark, line 8: Shouldn't the object actually be called `EvaluationCriterion`? Then we'll also pass a `List[EvaluationCriterion]` to a unit test upon creation. It's also a `List[int]` and not a `List[intS]` ;)

nucleus/modelci/eval_functions/base_eval_function.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    def __init__(self, eval_func_entry: EvalFunctionEntry):
+        self.eval_func_entry = eval_func_entry
+
+    @classmethod
+    @property
+    @abc.abstractmethod
+    def name(cls) -> str:
+        """Name to look for in the EvalFunctionDefinitions"""
```
pfmark, line 23: Would it also make sense to have annother abstract method which returns the eval function id `ef_...`? I think this would be a nicer interface for the user, rather than to got through the client and list them first, as we have knowledge about the specific one here (or rather in the child class)

pfmark (changes requested):
Few small comments. I really like the new interface. Let's user set up tests quickly and is not error prone anymore. It's a bit annoying that we need to add eval functions at that many places, but I think we can live with that for now and rather have a neat UX. 

Let's add an assertion that we need at least one eval function in order to create a unit test. 

Will also need to adapt the docstrings then, but we can do this in another PR once this is merged. I started with that already for other existing stuff.

tests/modelci/test_unit_test.py
pfmark: As long as it's a function it should be `get_...`, if it's `criteria` only it should become a property.

pfmark (approved):
lgtm, if checks pass ;)

tests/modelci/test_eval_function.py
gatli: Changed it to check for right instance type.

nucleus/modelci/unit_test.py
gatli: Removing!

nucleus/modelci/__init__.py
gatli: I've added an empty check.

gatli:
Currently blocked by backend deployment and bugfix https://github.com/scaleapi/scaleapi/pull/33371

nucleus/modelci/unit_test.py
phil-scale: I don't think this method is necessary or belongs in `unit_test.py` now.

sasha-scale (approved):
Let's make sure documentation is up to date with this change!

pfmark:
> Let's make sure documentation is up to date with this change!

Will follow in a second PR once this is merged!

PR #162, opened 2021-12-01
integration tests for public dataset

ardila:
Tests that we can export items from public datasets.

Will need to figure out how to handle the expired signed URL issue when we run into that.

### DIFF
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -564,7 +564,9 @@ class Dataset:
                     }
                 }
         """
-        response = self._client.dataitem_iloc(self.id, i)
+        response = self._client.make_request(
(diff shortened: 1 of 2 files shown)


nucleus/dataset.py
```diff
@@ -564,7 +564,9 @@ def iloc(self, i: int) -> dict:
                     }
                 }
         """
-        response = self._client.dataitem_iloc(self.id, i)
+        response = self._client.make_request(
```
song-william, line 567: any particular reason we don't use the existing `dataitem_iloc` anymore?

song-william (approved):

nucleus/dataset.py
ardila: Yes, it's deprecated.

PR #164, opened 2021-12-02
Support for default category taxonomy

cmpajot:
Adds API support for a default category taxonomy so that users don't have to explicitly specify a taxonomy ahead of time. Unit test included in PR and tested with code in scaleapi/scaleapi#33281.

[[sc-268730]](https://app.shortcut.com/scaleai/story/268730)

### DIFF
--- a/nucleus/annotation.py
+++ b/nucleus/annotation.py
@@ -515,16 +515,16 @@ class CategoryAnnotation(Annotation):
 
         category = CategoryAnnotation(
             label="dress",
-            taxonomy_name="clothing_type",
             reference_id="image_1",
+            taxonomy_name="clothing_type",
             metadata={"dress_color": "navy"}
         )
 
     Parameters:
         label (str): The label for this annotation.
-        taxonomy_name (str): The name of the taxonomy this annotation conforms to.
-          See :meth:`Dataset.add_taxonomy`.
         reference_id (str): User-defined ID of the image to which to apply this annotation.
+        taxonomy_name (Optional[str]): The name of the taxonomy this annotation conforms to.
+          See :meth:`Dataset.add_taxonomy`.
         metadata (Optional[Dict]): Arbitrary key/value dictionary of info to attach to this annotation.
           Strings, floats and ints are supported best by querying and insights
           features within Nucleus. For more details see our `metadata guide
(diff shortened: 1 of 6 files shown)


nucleus/prediction.py
```diff
@@ -365,8 +365,8 @@ class CategoryPrediction(CategoryAnnotation):
     def __init__(
         self,
         label: str,
-        taxonomy_name: str,
         reference_id: str,
+        taxonomy_name: Optional[str] = None,
```
ardila, line 369: Heads up that changing argument order can be a breaking change. A hack to get around this is to manually validate required things are passed in (i.e. reference id). Fine in this case though since it's a new API, just FYI.

nucleus/annotation.py
```diff
@@ -575,20 +577,22 @@ def __post_init__(self):
     def from_json(cls, payload: dict):
         return cls(
             labels=payload[LABELS_KEY],
-            taxonomy_name=payload[TAXONOMY_NAME_KEY],
+            taxonomy_name=payload.get(TAXONOMY_NAME_KEY, None),
```
ardila, line 580: If the default taxonomy name is "default" why can't users see this in their API objects?

ardila (approved):

nucleus/annotation.py
cmpajot: My view is that it's more elegant if we abstract away the whole concept of taxonomies in the case where users just use a default taxonomy but we could also expose the fact that this is using a default taxonomy under the hood. @rkaplan do you have an opinion on whether the default taxonomy should be exposed or hidden?

nucleus/annotation.py
```diff
@@ -532,8 +532,8 @@ class CategoryAnnotation(Annotation):
     """
 
     label: str
-    taxonomy_name: str
     reference_id: str
+    taxonomy_name: Optional[str] = None
```
rkaplan, line 536: Missing the ordering change in the docstring above

nucleus/annotation.py
rkaplan: Agree I think the default should ideally be hidden.

Imo, taxonomy names should feel like a slightly more advanced feature above the base case, which we expect to be more common, which is simple classification tasks. For someone walking through an MNIST upload tutorial, to me it feels like they should never need to understand taxonomy names. It's just a category label. Wdyt @ardila ?

nucleus/prediction.py
rkaplan: Doc string needs to be updated here as well to indicate taxonomy_name is optional

cmpajot: It actually will autoupdate here from the function arguments. I ended up specifying it for `CategoryAnnotation` because it wouldn't autopopulate in the docs when I initially wrote the docstrings. @drakejwong  do you have any idea why that was the case?

PR #165, opened 2021-12-03
Creating specific datasets for scenes

pfmark:
# Summary
We'll treat datasets that contain scenes differently in the frontend (additional tab) compared to those without scenes. Therefore we need to introduce an attribute for a dataset which specifies the type of dataset as a boolean (`is_scene`). This attribute needs to be set during dataset creation. 

Regarding the minor version, I assumed, we need to see if we merge [Gunnar's PR](https://github.com/scaleapi/nucleus-python-client/pull/159) first. So if this one would be merged first, it should be `0.5.0` instead. 

# How to test
This PR depends on the backend changes introduced in PR [#33106](https://github.com/scaleapi/scaleapi/pull/32750). To test it, do the following
- set `NUCLEUS_ENDPOINT = "http://localhost:3000/v1/nucleus"` in `nucleus/constants.py`
- Install this client from `pip install .` from source folder
- Check out `scaleapi` branch `mark-add-scene-type`
- yarn `dev-prodclone-backend`
- yarn `dev-prodclone-sfn`
- Create a dataset and upload a scene through the API. I can share a little script for that, let me know if you need sth.


tests/test_dataset.py
```diff
@@ -99,11 +99,28 @@ def make_dataset_items():
     return ds_items_with_metadata
 
 
-def test_dataset_create_and_delete(CLIENT):
+def test_dataset_create_and_delete_no_scene(CLIENT):
```
ardila, line 102: Shouldn't you also have tests for what effect this has on upload?

ardila:
Why are we making users explicitly set this rather than just inferring it from the data they send in?

ardila:
I'll LG for now since I assume these design discussions have alreaady ocurred and I don't want to be an obstacle to progress here. We do need to immediately update the guides for scene uploads concurrently with the release of any version of the client which includes this change.

nucleus/__init__.py
```diff
@@ -337,6 +339,7 @@ def create_dataset(
 
         Parameters:
             name: A human-readable name for the dataset.
+            is_scene: Boolean specifying if the dataset contains / will contain scenes. Cannot be changed at a later point in time.
```
ardila, line 342: Add more detail on what this will affect.

ardila (approved):

pfmark:
> I'll LG for now since I assume these design discussions have alreaady ocurred and I don't want to be an obstacle to progress here. We do need to immediately update the guides for scene uploads concurrently with the release of any version of the client which includes this change.

Yup, I had it implicitly in the beginning but @song-william saw DB issues there. Therefore it's explicit now. I'm also fine having the default value `True` now. This will soon end up in an additional empty tab in the frontend then though.

jean-lucas:
I also agree that the `is_scene` should be set programmatically, without the user input. Two reasons i can think of why:
- will break current workflows
- user may not know what a scene is

pfmark:
> I also agree that the is_scene should be set programmatically, without the user input. Two reasons i can think of why:

See discussion [here](https://github.com/scaleapi/scaleapi/pull/33106), it shouldn't change for a dataset.

> will break current workflows

Not really, we just need to update the docs accordingly. 

> user may not know what a scene is

I'm pretty sure they wouldn't upload a scene in that case. 

As I mentioned above: 
- Either we set the default value to `False` + careful update of the docs (preferred)
- Set default to `True` and live with the empty scene tab in the frontend

nucleus/dataset.py
```diff
@@ -118,6 +118,14 @@ def name(self) -> str:
             )["name"]
         return self._name
 
+    @property
+    def is_scene(self) -> bool:
```
jean-lucas, line 122:
```
dataset = client.create_dataset('Test PR 165', is_scene=True)
dataset.is_scene
```

Gives the error:
```
NucleusAPIError: Your client is on version 0.3.1. If you have not recently done so, please make sure you have updated to the latest version of the client by running pip install --upgrade scale-nucleus
Tried to get http://localhost:3000/v1/nucleus/dataset/ds_c6n3ysk5e8cp828vqnj0/is_scene, but received 404: Not Found.
The detailed error is:
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /v1/nucleus/dataset/ds_c6n3ysk5e8cp828vqnj0/is_scene</pre>
</body>
</html>
```

nucleus/dataset.py
```diff
@@ -118,6 +118,14 @@ def name(self) -> str:
...
+        """If the dataset can contain scenes or not."""
+        response = self._client.make_request(
+            {}, f"dataset/{self.id}/is_scene", requests.get
+        )["is_scene"]
```
jean-lucas, line 126:
```suggestion
        )[DATASET_IS_SCENE_KEY]
```

jean-lucas (changes requested):
You may also need to update the patch or minor version of the pip package before merging

nucleus/dataset.py
jean-lucas: Maybe need to add the import. Haven't tested suggestion

nucleus/dataset.py
jean-lucas: it works now after pulling

tests/test_dataset.py
pfmark: I'm testing this properly in the backend already. You think I should test it here again?

nucleus/dataset.py
jean-lucas:
```
/usr/local/lib/python3.9/site-packages/nucleus/dataset.py in is_scene(self)
    124         response = self._client.make_request(
    125             {}, f"dataset/{self.id}/is_scene", requests.get
--> 126         )[DATASET_IS_SCENE_KEY]
    127         return response
    128 

NameError: name 'DATASET_IS_SCENE_KEY' is not defined
```

nucleus/__init__.py
```diff
@@ -329,6 +330,7 @@ def create_dataset_from_project(
     def create_dataset(
```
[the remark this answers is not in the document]
pfmark: Should be resolved now, right?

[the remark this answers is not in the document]
pfmark: Okay will do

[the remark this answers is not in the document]
sasha-scale: I think this is resolved now

nucleus/__init__.py
```diff
@@ -354,9 +363,17 @@ def create_dataset(
         Returns:
             :class:`Dataset`: The newly created Nucleus dataset as an object.
         """
+        warnings.warn(
+            "The default create_dataset('dataset_name', ...) method will be deprecated soon. "
```
sasha-scale, line 367:
```suggestion
            "The default create_dataset('dataset_name', ...) method without the is_scene parameter will be deprecated soon in favor of providing the is_scene parameter explicitly. "
```

pyproject.toml
```diff
@@ -21,7 +21,7 @@ exclude = '''
 
 [tool.poetry]
 name = "scale-nucleus"
-version = "0.3.1"
+version = "0.4.0"
```
sasha-scale, line 24: we just incremented this for Model CI 😢 
```suggestion
version = "0.4.1"
```

sasha-scale (approved):
lgtm barring comments

tests/test_dataset.py
pfmark: It's both covered in the backend and here now.

jean-lucas (approved):
LGTM 👍

PR #177, opened 2021-12-13
Clean up deprecations warnings in test suite and add parallel workers

gatli:
Cleans up some of the deprecation warnings while running the tests. Some logic is still kind of circular so we'll need to take another pass.

I've also added parallel runs for the tests so they are currently passing (on the same machine) in 10 mins vs 20 before.


nucleus/__init__.py
```diff
@@ -659,11 +659,16 @@ def dataitem_ref_id(self, dataset_id: str, reference_id: str):
...
     @deprecated("Prefer calling Dataset.predictions_refloc instead.")
     @sanitize_string_args
-    def predictions_ref_id(self, model_run_id: str, ref_id: str):
+    def predictions_ref_id(
+        self, model_run_id: str, ref_id: str, dataset_id: Optional[str] = None
+    ):
+        if dataset_id:
+            raise RuntimeError(
+                "Need to pass a dataset id. Or use Dataset.predictions_refloc."
+            )
```
gatli, line 668: I'm not quite sure how this worked before but I just added a optional dataset_id such that this works here but trips up existing usage and points them to the the non-deprecated method.

pyproject.toml
```diff
@@ -57,6 +57,7 @@ Sphinx = "^4.2.0"
 sphinx-autobuild = "^2021.3.14"
 furo = "^2021.10.9"
 sphinx-autoapi = "^1.8.4"
+pytest-xdist = "^2.5.0"
```
gatli, line 60: Allows parallel local runs with `pytest -n auto`

nucleus/__init__.py
ardila: This is fine :)

pyproject.toml
ardila: very nice

ardila:
Need to fix merge conflicts, but looks good to me

ardila (approved):

sasha-scale (approved):

PR #179, opened 2021-12-15
Fix Scene metadata update workflow: get scene followed by re-upload should be seamless

sasha-scale:
Corresponds to: https://github.com/scaleapi/scaleapi/pull/33731

The workflow of scene.get -> scene.upload was broken because the object returned by get was instantiated as Scene instead of lidar scene..  Added pytest integration test to handle this case as well as the update flag causing update on conflict behavior, which was previously missing.

### DIFF
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -1016,7 +1016,7 @@ class Dataset:
             :class:`Scene<LidarScene>`: A scene object containing frames, which
             in turn contain pointcloud or image items.
         """
-        return Scene.from_json(
+        return LidarScene.from_json(
             self._client.make_request(
                 payload=None,
                 route=f"dataset/{self.id}/scene/{reference_id}",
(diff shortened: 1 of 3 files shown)


nucleus/dataset.py
```diff
@@ -1016,7 +1016,7 @@ def get_scene(self, reference_id: str) -> Scene:
             :class:`Scene<LidarScene>`: A scene object containing frames, which
             in turn contain pointcloud or image items.
         """
-        return Scene.from_json(
+        return LidarScene.from_json(
```
ardila, line 1019: Should the backend return something so we know what type of scene to instantiate? This change means we can't get anything EXCEPT lidar scenes, which is not everything right?

ardila (approved):
Approve for now, maybe file a ticket to address the issue I raised if you agree?

nucleus/dataset.py
sasha-scale: Right now the only possible type of scene is LidarScene, and we intend to extend this to support videos later on.  Inclined to keep it this way for now, and revisit when we implement general video scene support

PR #181, opened 2021-12-16
Speed up ModelCI tests and add annotations as are required by backend

gatli:
Refactored the tests such that they fit the new requirements on the backend. Now we require box annotations to be associated with the dataset items in the slice. This is implemented in https://github.com/scaleapi/scaleapi/pull/33821.

I used the opportunity to refactor the model ci tests to reuse slow to set up test fixture and making the scope of them `module` (which makes them live longer).

Test time for Model CI goes from around 40s down 10s on my machine (except for the new integration test that is tracking an error) by using module scoped fixtures.

### DIFF
--- a/nucleus/__init__.py
+++ b/nucleus/__init__.py
@@ -35,7 +35,6 @@ __all__ = [
 ]
 
 import os
-import time
 import warnings
 from typing import Dict, List, Optional, Sequence, Union
 
@@ -102,6 +101,7 @@ from .errors import (
     DatasetItemRetrievalError,
     ModelCreationError,
     ModelRunCreationError,
+    NoAPIKey,
     NotFoundError,
     NucleusAPIError,
 )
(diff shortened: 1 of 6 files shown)


nucleus/__init__.py
```diff
@@ -935,3 +934,15 @@ def handle_bad_response(
...
+
+    def _set_api_key(self, api_key):
+        """Fetch API key from environment variable NUCLEUS_API_KEY if not set"""
+        api_key = (
+            api_key if api_key else os.environ.get("NUCLEUS_API_KEY", None)
+        )
+        if api_key is None:
+            raise NoAPIKey(
+                "You need to pass an API key to the NucleusClient or set the environment variable NUCLEUS_API_KEY"
+            )
```
phil-scale, line 946: Could you make this error the default message of the `NoAPIKey` class?

phil-scale (approved):
Approving with minor comment - thanks for fixing this!

PR #185, opened 2022-01-03
Add basic polygon metrics to nucleus repo

phil-scale:
Adds IoU, precision, and recall metrics to the Nucleus repository. Creates framework to support more general metrics.

**Note:** adding the `Shapely` package also adds a dependency on the `geos` library, which can be installed in MacOS using `brew install geos`.

Design doc here: https://docs.google.com/document/d/1uOjZQ9wRMdpsOaYN_w6eTqPvQu0Nwtl5wRNj0HkL0EU/edit#heading=h.raendxyxrl42
[ch318548](https://app.shortcut.com/scaleai/story/318548)
[ch303550](https://app.shortcut.com/scaleai/story/303550)

### DIFF
--- a/nucleus/metrics/base.py
+++ b/nucleus/metrics/base.py
@@ -21,7 +21,7 @@ class MetricResult:
     """
 
     value: float
-    weight: float = 1
+    weight: float = 1.0
 
     @staticmethod
     def aggregate(results: Iterable["MetricResult"]) -> "MetricResult":
--- a/nucleus/metrics/base.py
+++ b/nucleus/metrics/base.py
@@ -11,7 +11,7 @@ AnnotationOrPredictionList = Union[List[Annotation], List[Prediction]]
 
 @dataclass
 class MetricResult:
-    """An Metric Result contains the value of an evaluation, as well as its weight.
+    """A Metric Result contains the value of an evaluation, as well as its weight.
     The weight is useful when aggregating metrics where each dataset item may hold a
     different relative weight. For example, when calculating precision over a dataset,
     the denominator of the precision is the number of annotations, and therefore the weight
--- a/nucleus/metrics/base.py
+++ b/nucleus/metrics/base.py
@@ -1,12 +1,10 @@
 import sys
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Iterable, List, Union
+from typing import Iterable
 
-from nucleus.annotation import Annotation
-from nucleus.prediction import Prediction
-
-AnnotationOrPredictionList = Union[List[Annotation], List[Prediction]]
+from nucleus.annotation import AnnotationList
+from nucleus.prediction import PredictionList
 
 
 @dataclass
@@ -25,47 +23,67 @@ class MetricResult:
     value: float
     weight: float = 1
 
-    @classmethod
-    def aggregate(cls, results: Iterable["MetricResult"]) -> "MetricResult":
+    @staticmethod
+    def aggregate(results: Iterable["MetricResult"]) -> "MetricResult":
         """Aggregates results using a weighted average."""
         results = list(filter(lambda x: x.weight != 0, results))
         total_weight = sum([result.weight for result in results])
         total_value = sum([result.value * result.weight for result in results])
         value = total_value / max(total_weight, sys.float_info.epsilon)
-        return cls(value, total_weight)
+        return MetricResult(value, total_weight)
 
 
 class Metric(ABC):
-    """Abstract class to define a Metric."""
+    """Abstract class for defining a metric, which takes a list of annotations
+    and predictions and returns a scalar.
+
+    To create a new concrete Metric, override the `__call__` function
+    with logic to define a metric between annotations and predictions. ::
+
+        from nucleus import BoxAnnotation, CuboidPrediction, Point3D
+        from nucleus.annotation import AnnotationList
+        from nucleus.prediction import PredictionList
+        from nucleus.metrics import Metric, MetricResult
+        from nucleus.metrics.polygon_utils import BoxOrPolygonAnnotation, BoxOrPolygonPrediction
+
+        class MyMetric(Metric):
+            def __call__(
+                self, annotations: AnnotationList, predictions: PredictionList
+            ) -> MetricResult:
+                value = (len(annotations) - len(predictions)) ** 2
+                weight = len(annotations)
+                return MetricResult(value, weight)
+
+        box = BoxAnnotation(
+            label="car",
+            x=0,
+            y=0,
+            width=10,
+            height=10,
+            reference_id="image_1",
+            annotation_id="image_1_car_box_1",
+            metadata={"vehicle_color": "red"}
+        )
+
+        cuboid = CuboidPrediction(
+            label="car",
+            position=Point3D(100, 100, 10),
+            dimensions=Point3D(5, 10, 5),
+            yaw=0,
+            reference_id="pointcloud_1",
+            confidence=0.8,
+            annotation_id="pointcloud_1_car_cuboid_1",
+            metadata={"vehicle_color": "green"}
+        )
+
+        metric = MyMetric()
+        annotations = AnnotationList(box_annotations=[box])
+        predictions = PredictionList(cuboid_predictions=[cuboid])
+        metric(annotations, predictions)
+    """
 
     @abstractmethod
     def __call__(
-        self, annotations: List[Annotation], predictions: List[Prediction]
+        self, annotations: AnnotationList, predictions: PredictionList
     ) -> MetricResult:
         """A metric must override this method and return a metric result, given annotations and predictions."""
-
-
-class Filter(ABC):
-    """Abstract class to define an Filter"""
-
-    @abstractmethod
-    def __call__(
-        self, annotations: AnnotationOrPredictionList
-    ) -> AnnotationOrPredictionList:
-        """A Filter must override this method."""
-
-
-class AnnotationFilter(ABC):
-    """Abstract class to define an Filter"""
-
-    @abstractmethod
-    def __call__(self, annotations: List[Annotation]) -> List[Annotation]:
-        """An Annotation Filter must override this method."""
-
-
-class PredictionFilter(ABC):
-    """Abstract class to define an Prediction Filter"""
-
-    @abstractmethod
-    def __call__(self, predictions: List[Prediction]) -> List[Prediction]:
-        """An Prediction Filter must override this method."""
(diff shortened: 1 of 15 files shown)


phil-scale:
Note: this bumps the python requirement to `>= 3.7` in order to support `scipy` - is that ok?

nucleus/metrics/base.py
```diff
@@ -0,0 +1,71 @@
...
+        """Aggregates results using a weighted average."""
+        results = list(filter(lambda x: x.weight != 0, results))
+        total_weight = sum([result.weight for result in results])
+        total_value = sum([result.value * result.weight for result in results])
+        value = total_value / max(total_weight, sys.float_info.epsilon)
+        return cls(value, total_weight)
+
+
+class Metric(ABC):
+    """Abstract class to define a Metric."""
```
gatli, line 39: The docstring isn't saying anything else than the preceding line currently. Here would be a good place to explain the definition of a Metric.

nucleus/metrics/base.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @abstractmethod
+    def __call__(
+        self, annotations: List[Annotation], predictions: List[Prediction]
+    ) -> MetricResult:
```
gatli, line 44: ❤️ It's nice that you already wrap the results in a class. Makes things more maintainable in the future 🙂

nucleus/metrics/base.py
```diff
@@ -0,0 +1,71 @@
...
+        """A metric must override this method and return a metric result, given annotations and predictions."""
+
+
+class Filter(ABC):
+    """Abstract class to define an Filter"""
```
gatli, line 49: I'm not sure what a filter is. Could you elaborate in this docstring?

nucleus/metrics/errors.py
```diff
@@ -0,0 +1,7 @@
+class PolygonAnnotationTypeError(Exception):
+    def __init__(
+        self,
+        message="Annotation was expected to be type BoxAnnotation or PolygonAnnotation.",
```
gatli, line 4: [nit]
```suggestion
        message="Annotation was expected to be of type 'BoxAnnotation' or 'PolygonAnnotation'.",
```

nucleus/metrics/filters.py
```diff
@@ -0,0 +1,48 @@
+from typing import List, Tuple
+
+from nucleus.annotation import Annotation
+from nucleus.prediction import Prediction
+
+from .base import AnnotationOrPredictionList, Filter, PredictionFilter
+from .polygon_utils import BoxOrPolygonAnnotation, _polygon_annotation_to_shape
```
gatli, line 7: I'd convert the second method to public if you are using it outside of the module.

nucleus/metrics/polygon_metrics.py
```diff
@@ -0,0 +1,212 @@
...
+            y=0,
+            width=10,
+            height=10,
+            reference_id="image_1",
+            annotation_id="image_1_car_box_1",
+            metadata={"vehicle_color": "red"}
+        )
+
+        metric = MyPolygonMetric()
+        metric([box], [box])
```
gatli, line 59: ❤️

pyproject.toml
```diff
@@ -21,7 +21,7 @@ exclude = '''
 
 [tool.poetry]
 name = "scale-nucleus"
-version = "0.4.2"
+version = "0.4.3"
```
gatli, line 24: This is adding new features so semantic versioning says to bump minor 🙂 
```suggestion
version = "0.5.0"
```

pyproject.toml
```diff
@@ -32,7 +32,7 @@ documentation = "https://dashboard.scale.com/nucleus/docs/api"
 packages = [{include="nucleus"}]
 
 [tool.poetry.dependencies]
-python = "^3.6.2"
+python = ">=3.7,<3.11"
```
gatli, line 35: I think this is OK 
> Because Python 3.6 extended support on Windows and Linux will end on 23 December 2021, 
as long as we make sure to signal to users that they need to upgrade the Python version.

I'm wondering how pip deals with this. If people are running Python 3.6 will pip just refuse to list newer version and upgrade? We need to add some sort of message hinting people to upgrade to Python3.7 +.

We might actually have to list the depends as ">=3.6" and the dev depends as ">=3.7" and add a error message in`__init__` for 3.6 hinting to upgrade.

gatli (changes requested):
Very clear code to read 🙂 

I've requested changes since we need to be sure that Python 3.6 users see a warning message to upgrade. Until then I'm not comfortable rolling this out since I know that python3.6 is default in some widely used distros.

tests/metrics/test_polygon_metrics.py
```diff
@@ -0,0 +1,20 @@
+from nucleus.metrics import PolygonIOU, PolygonPrecision, PolygonRecall
+from tests.metrics.helpers import (
+    TEST_BOX_ANNOTATIONS,
+    TEST_CONVEX_POLYGON_ANNOTATIONS,
+    perfect_match_test_wrapper,
+)
+
+
+def test_perfect_match_box_metrics():
```
ardila, line 9: Seems like you should test a non-perfect match just to make sure logic is correct?

ardila: Also these tests should be split up so that things can fail separately.

I would recommend writing a single test called identical_values_result_in_a_value_of_1

And then parametrize it

https://docs.pytest.org/en/latest/how-to/parametrize.html

nucleus/metrics/base.py
ardila: +1

nucleus/metrics/base.py
ardila: Yeah I think you should include some explanation of the framework here, what is required to be defined, plus some example code of how to use this base class (can use made-up functions that aren't defined in this snippet)

nucleus/metrics/base.py
phil-scale: Removed the filter. Wasn't super necessary...

tests/metrics/test_polygon_metrics.py
phil-scale: Yes, will add more tests. Thanks for the parametrize suggestion, did not know of this!

pyproject.toml
phil-scale: Actually I resolved this by making scipy/numpy part of dev requirements and having those specific ones require higher python versions :)

phil-scale:
> Very clear code to read 🙂 
> 
> 
> 
> I've requested changes since we need to be sure that Python 3.6 users see a warning message to upgrade. Until then I'm not comfortable rolling this out since I know that python3.6 is default in some widely used distros.

Moved the new dependencies to dev only, with the Python requirement applied specifically to those dependencies.

nucleus/metrics/base.py
```diff
@@ -0,0 +1,89 @@
...
+    def aggregate(results: Iterable["MetricResult"]) -> "MetricResult":
+        """Aggregates results using a weighted average."""
+        results = list(filter(lambda x: x.weight != 0, results))
+        total_weight = sum([result.weight for result in results])
+        total_value = sum([result.value * result.weight for result in results])
+        value = total_value / max(total_weight, sys.float_info.epsilon)
+        return MetricResult(value, total_weight)
+
+
+class Metric(ABC):
```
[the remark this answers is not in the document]
yixu34: O nm you're already on this PR

nucleus/metrics/polygon_utils.py
```diff
@@ -0,0 +1,230 @@
...
+
+from nucleus.annotation import BoxAnnotation, PolygonAnnotation
+from nucleus.prediction import BoxPrediction, PolygonPrediction
+
+from .base import MetricResult
+from .errors import PolygonAnnotationTypeError
+
+BoxOrPolygonPrediction = Union[BoxPrediction, PolygonPrediction]
+BoxOrPolygonAnnotation = Union[
+    BoxAnnotation, PolygonAnnotation, BoxOrPolygonPrediction
```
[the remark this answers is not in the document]
phil-scale: Yeah it's tough because `Prediction` actually inherits from `Annotation` but the `List` template does not support inheritance.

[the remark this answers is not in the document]
phil-scale: On second thought, I could play around with a `TypeVar` here instead.

nucleus/metrics/polygon_metrics.py
```diff
@@ -0,0 +1,213 @@
...
+        return MetricResult(avg_iou, weight)
+
+
+class PolygonPrecision(PolygonMetric):
+    """Calculates the precision between box or polygon annotations and predictions."""
+
+    def __init__(
+        self,
+        enforce_label_match: bool = False,
+        iou_threshold: float = 0.5,
```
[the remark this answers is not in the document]
phil-scale: The idea is that in Nucleus and the most commonly reported metrics, it's usually `0.5` by default. Also if we're exposing these evaluation functions to users, we would want users to know the default values that are actually used in Model CI.

nucleus/metrics/filters.py
```diff
@@ -0,0 +1,42 @@
...
+
+def confidence_filter(
+    predictions: PredictionList, min_confidence: float
+) -> PredictionList:
+    predictions_copy = PredictionList()
+    filter_fn = (
+        lambda prediction: not hasattr(prediction, "confidence")
+        or prediction.confidence >= min_confidence
+    )
+    predictions_copy.box_predictions = list(
```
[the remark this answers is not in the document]
phil-scale: Heh I can do this with `__dict__` - maybe a little hacky but I think it's pretty understandable still.

nucleus/metrics/polygon_metrics.py
```diff
@@ -0,0 +1,213 @@
...
+        return result
+
+
+class PolygonIOU(PolygonMetric):
+    """Calculates the average IOU between box or polygon annotations and predictions."""
+
+    def __init__(
+        self,
+        enforce_label_match: bool = False,
+        iou_threshold: float = 0.0,
```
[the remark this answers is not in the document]
phil-scale: My thinking is that we can expose information about what defaults are used in Model CI in two ways:
1) Currently, we can keep these defaults in code so that users who want to see what the values are can find them directly here as defaults.
2) Once the UI shows the source code of evaluation functions, we can get rid of the defaults here and instead show them in the source code (e.g. `PolygonIOU(enforce_label_match=True, iou_threshold=0.2, confidence_threshold=0.5)`).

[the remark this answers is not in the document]
phil-scale: In that case, I'll add a TODO here in `polygon_metrics.py` to remove the defaults once we can surface these defaults more directly to the user.

gatli (approved):
LGTM 👍 🔥

gatli:
Rebased on top of master ...

PR #188, opened 2022-01-05
use canary, not prod endpoint

drakejwong:
QoL change so we don't have to wait for hotfix/daily deploy to update the backend.

Won't bump version or publicize the change since this is internal


gatli (changes requested):
This allows us to push an update that is broken until the next releaase. This could mean broken features for a day under normal circumstances. I think thats a rather long window.

I think we should rather override the NUCLEUS_ENDPOINT in the daily circleci tests such that we might get a head start on a bug deployed to canary. I think this was @ardila intent when we talked about changing tests to the canary.

If we want to run canary tests for each PR then I think we should duplicate the test suite with different params. That way the tests aren't green until production is updated but we still get feedback from the canary that the code is correct.

drakejwong:
> This allows us to push an update that is broken until the next releaase. This could mean broken features for a day under normal circumstances. I think thats a rather long window.
> 
> I think we should rather override the NUCLEUS_ENDPOINT in the daily circleci tests such that we might get a head start on a bug deployed to canary. I think this was @ardila intent when we talked about changing tests to the canary.
> 
> If we want to run canary tests for each PR then I think we should duplicate the test suite with different params. That way the tests aren't green until production is updated but we still get feedback from the canary that the code is correct.

Yes totally agree, I did not aptly consider the production dependencies that exist after merging. I think the duplicated approach sounds good — canary to get a head start on validating code before deploy + prod to actually block the PR until backend is deployed. In which case we honestly might be better off leaving things as is...

PR #194, opened 2022-01-11
Try to cache dependencies v2

gatli:


gatli:
Got the caching to work but it seems to have little to no effect. The second run had a cache hit but the installation step took exactly as long of a time, around 1:30.

PR #196, opened 2022-01-11
Make dependencies work for python 3.6

ardila:
* Keep testing on python 3.6
* Move dependencies out of dev 
* Change the versions of dependencies we require so that they can be installed on python 3.6


phil-scale:
Should we also publish onto pypi on the `cimg/python:3.6` image then?

phil-scale (approved):
LGTM

PR #198, opened 2022-01-12
Explain EvaluationCriterion "magic" and documenting Pydantic usage

gatli:
Adds documentation around `EvaluationCriterion` overloading and the reasoning for it as requested.

Also adds a section to `README.md` on the usage of Pydantic for development.

If there is anything that needs further clarification or is uncertain then please point it out 🙂 !

## On the usage of Pydantic
Here's a good comparison of `dataclasses` vs `pydantic` vs `attrs`: https://jackmckew.dev/dataclasses-vs-attrs-vs-pydantic.html

- I think `pydantic` is the best fit here since it provides very good error message and validation facilities that will help users self serve and report problems.
- `attrs` seems like a good fit if we need more customizable structuring and destructuring but the error messages seem to be a lot worse according to https://threeofwands.com/why-i-use-attrs-instead-of-pydantic/


nucleus/modelci/README.md
```diff
@@ -0,0 +1,45 @@
+# Notes on defining Evaluation Criteria for a test
```
ardila, line 1: This is great documentation, but I think here it may not be very discoverable. Can you include the stuff under this header in the class docstring for EvaluationCriterion instead? I think users will find it more easily there.

ardila: I think both headers can go under the class docstring for evaluation criterion. This way it will show up in our API docs.

I think users will either look at our "how to guides" or our API docs in most cases.

ardila (approved):

nucleus/modelci/README.md
phil-scale: +1 I think this README is fantastic and should stay. Maybe we could add a summary of this explanation to the `EvaluationCriterion` class docstring.

phil-scale (approved):
This is great, thanks!

README.md
```diff
@@ -62,6 +62,90 @@ poetry run pytest tests/test_dataset.py
 poetry run pytest -m "not integration"
 ```
 
+## Pydantic Models
```
gatli, line 65: @ardila @phil-scale somehow forgot to push this. Could you take a look at this as well?

sasha-scale (approved):

PR #199, opened 2022-01-12
Add dataset.delete_scene

gatli:
Adds Scene deletion through the SDK.

Still waiting for the endpoint to be deployed!

### DIFF
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -742,7 +742,14 @@ class Dataset:
 
     @sanitize_string_args
     def delete_scene(self, reference_id: str):
-        """ Deletes a Scene associated with the Dataset"""
+        """Deletes a Scene associated with the Dataset
+
+        All items, annotations and predictions associated with the scene will be
+        deleted as well.
+
+        Parameters:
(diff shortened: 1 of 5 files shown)


nucleus/dataset.py
```diff
@@ -738,6 +740,11 @@ def delete_item(self, reference_id: str) -> dict:
             requests.delete,
         )
 
+    @sanitize_string_args
+    def delete_scene(self, reference_id: str):
```
pfmark, line 744: Wondering if it makes more sense to just have a `delete_scenes(self, reference_ids: List[str])` method. We should also adapt the interface or at least add this method for the other data types to make it easier for the user and interate through the scenes internally rather then on the user side. Wdyt @gatli ?

pfmark: Can also do this in a second PR actually. So let's go ahead and merge once the endpoint is deployed.

pfmark (approved):
lgtm

nucleus/dataset.py
gatli: Agreed, that is probably a more natural interface. Let's add it to shortcut!

PR #200, opened 2022-01-14
Add AP and MAP to polygon metrics

phil-scale:
Adds average precision and mean average precision for boxes/polygons.

Adds more informative docstring for `PolygonIOU`, `PolygonPrecision`, and `PolygonRecall`.


gatli (approved):
LGTM 👍 Sorry about the late review. Seems to have slipped between the cracks.

PR #201, opened 2022-01-17
Make shapely installation optional via metrics flag

gatli:
Shapely seems to be a nice package for polygon metrics. It was a little involved for me to switch it out right now so I opted to make the dependency optional. It can now be installed with:

`pip install scale-nucleus[metrics]`

or locally

`poetry install -E metrics`

By default the shapely package is not installed. 

I think this is a fine solution for now, we can choose to remove the optional dependency in the future and change out `Shapely` for something else.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,10 +17,10 @@ jobs:
       - run:
           name: Install Environment Dependencies
           command: | # install dependencies
-            apt-get -y install curl
-            pip install --upgrade pip 
+            apt-get -y install curl libgeos-dev
+            pip install --upgrade pip
             pip install poetry
-            poetry install
+            poetry install -E metrics
 
       - run:
           name: Black Formatting Check # Only validation, without re-formatting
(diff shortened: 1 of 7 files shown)


phil-scale:
I also looked for the longest time for a package that could handle polygons (especially concave) properly and couldn't find a single one in Python.

I think this makes sense to me, though I wonder if we'd also make `numpy` and `scipy` optional as well then, since they're only used for `metrics` right now as well.

pfmark:
I'm fine with that solution. It's just important that we make it as transparent as possible, that a user would need to install with `pip install scale-nucleus[metrics]`. Let's also add it to the docstring of the `PolygonMetric` in order to have it in the API docs later.

ardila:
We are making a _very_ big change here (how users install the package) for a very small reason (calculating IOU metrics for polygons)

gatli:
> We are making a _very_ big change here (how users install the package) for a very small reason (calculating IOU metrics for polygons)

The change is very easily revertable and currently touches no users (nobody is using metrics locally). It also un-blocks rolling out the CLI.

Promise to re-address in the near future!

ardila:
Seems like we are closing in favor of removing the dependency. I'm going to close it for now since github bot keeps pinging me.

PR #202, opened 2022-01-18
Add Categorization F1

gatli:
Adds Categorization F1 score.

- I've adapted the `aggregate` methods to be defined by the metrics themselves instead of as a `staticmethod` of `MetricResult` that allows for more intricate aggregation methods than just weighing scalars like f1 score.

I'll follow up with a change to the celery service such that it adheres to the new interfaces.

This is the simplest way to get things up and running. More adaptation needs to be done to be able to highlight item level results.


phil-scale:
Ah I think the F1 score can actually be aggregated using the current weighted aggregation. Since F1 = 2 TP / (# predictions + # annotations), you would simply use the weight (# predictions + # annotations).

nucleus/metrics/categorization_metrics.py
```diff
@@ -0,0 +1,102 @@
...
+        ],  # TODO(gunnar): List to conform with other APIs or single instance?
+        predictions: List[CategoryPrediction],
+    ) -> CategorizationResult:
+        # TODO(gunnar): Allow passing multiple predictions and selecting highest confidence? Allows us to show next
+        #  contender. Are top-5 scores something that we care about?
+        # Main evaluation function that subclasses must override.
+        pass
+
+    @abstractmethod
+    def aggregate(self, results: List[CategorizationResult]) -> ScalarResult:  # type: ignore[override]
```
phil-scale, line 58: I wonder if you could get past typing errors with TypeVars here that are restricted to subclasses of `MetricResult`

tests/metrics/test_categorization_metrics.py
```diff
@@ -0,0 +1,84 @@
...
+        CategoryPrediction(label="sth_else", reference_id=ann.reference_id)
+        for ann in annotations
+    ]
+    metric = CategorizationF1()
+    results = []
+    for ann, pred in zip(annotations, predictions):
+        results.append(metric.eval([ann], [pred]))
+
+    assert results
+    aggregate_result = metric.aggregate(results)
```
phil-scale, line 40: I think it's important to be able to call the metric directly and use the `__call__` interface if possible so that the metrics all have the same interface.

nucleus/metrics/categorization_metrics.py
```diff
@@ -0,0 +1,102 @@
...
+        pass
+
+    def __call__(
```
phil-scale, line 61: Is this function used or tested?

nucleus/metrics/categorization_metrics.py
```diff
@@ -0,0 +1,102 @@
...
+            annotation=annotations[0], prediction=predictions[0]
+        )
+
+    def aggregate(self, results: List[CategorizationResult]) -> ScalarResult:  # type: ignore[override]
+        gt = []
+        predicted = []
+        for result in results:
+            gt.append(result.annotation.label)
+            predicted.append(result.prediction.label)
+        # TODO(gunnar): Support choice of averaging method
```
phil-scale, line 100: Maybe make this in the constructor?

nucleus/metrics/categorization_metrics.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+        return result
+
+
+class CategorizationF1(CategorizationMetric):
+    """Evaluation method that matches categories and returns a CategorizationF1Result that aggregates to the F1 score"""
+
+    def eval(
+        self,
+        annotations: List[CategoryAnnotation],
```
phil-scale, line 81: It seems a little strange to me to restrict this list to one element. The eval method here seems a little strange to begin with though

nucleus/metrics/categorization_metrics.py
gatli: Hmm, I'll have to try it out. Wasn't exactly in a `mypy` wrangling mood. Good suggestion! 🙂

tests/metrics/test_categorization_metrics.py
gatli: Will change!

nucleus/metrics/categorization_metrics.py
gatli: I'll add it there. This was more of a reminder that we need to add a way to surface these to the user 🙂

nucleus/metrics/categorization_metrics.py
gatli: Adapted this to support multiple taxonomies.

nucleus/metrics/categorization_metrics.py
gatli: Now it is 🙂

nucleus/metrics/categorization_metrics.py
```diff
@@ -0,0 +1,158 @@
...
+    predictions: List[CategoryPrediction]
+
+    @property
+    def value(self):
+        annotation_labels = to_taxonomy_labels(self.annotations)
+        prediction_labels = to_taxonomy_labels(self.predictions)
+
+        # TODO: Change task.py interface such that we can return label matching
+        # NOTE: Returning 1 if all taxonomy labels match else 0
+        return 1 if annotation_labels.difference(prediction_labels) == 0 else 0
```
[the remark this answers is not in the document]
gatli: This is still just the result shown in the frontend. The F1 score currently take into account each taxonomy individually into the calculation

So:
If you have a yellow car the predictions `type-> bus` and `color->yellow` will affect the F1 score individually. Even though we have a wrong type the `color->yellow` will push the F1 up.

nucleus/__init__.py
```diff
@@ -389,13 +389,14 @@ def create_dataset(
...
             :class:`Dataset`: The newly created Nucleus dataset as an object.
         """
-        warnings.warn(
-            "The default create_dataset('dataset_name', ...) method without the is_scene parameter will be deprecated soon in favor of providing the is_scene parameter explicitly. "
-            "Please make sure to create a dataset with either create_dataset('dataset_name', is_scene=False, ...) to upload "
-            "DatasetItems or create_dataset('dataset_name', is_scene=True, ...) to upload "
-            "LidarScenes.",
-            DeprecationWarning,
-        )
+        if is_scene is None:
```
gatli, line 392: 🥷 fix, this was causing `create_dataset` to be deprecated

nucleus/metrics/base.py
```diff
@@ -1,14 +1,18 @@
...
 
 
+class MetricResult(ABC):
+    """Base MetricResult class"""
+
+
 @dataclass
-class MetricResult:
+class ScalarResult(MetricResult):
     """A Metric Result contains the value of an evaluation, as well as its weight.
```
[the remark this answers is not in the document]
gatli: 👍

nucleus/metrics/base.py
```diff
@@ -87,3 +91,7 @@ def __call__(
         self, annotations: AnnotationList, predictions: PredictionList
     ) -> MetricResult:
         """A metric must override this method and return a metric result, given annotations and predictions."""
+
+    @abstractmethod
+    def aggregate_score(self, results: List[MetricResult]) -> ScalarResult:
+        """A metric must define how to aggregate results from single items to a single ScalarResult."""
```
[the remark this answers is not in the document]
gatli: Added one with R2 score.

gatli (commented):
I've addressed your comments, would be good to get this into master ASAP 🙂

PR #107, opened 2021-08-24
[Do not merge] Testing inference

seanshi-scale:
Companion: https://github.com/scaleapi/models/pull/1847
Superseded by https://github.com/scaleapi/nucleus-python-client/pull/203

PR #203, opened 2022-01-19
HMI Python client

seanshi-scale:
Supersedes https://github.com/scaleapi/nucleus-python-client/pull/107/commits


nucleus/deploy/find_packages.py
```diff
@@ -0,0 +1,330 @@
+#!/usr/bin/env python
```
seanshi-scale, line 1: Copied from `models/demos/find_packages.py` with some changes

yixu34 (commented):
Overall looks ok for now, though one thing I would like for us to resolve **before** the demo is refactoring the Endpoint class into sync/async variants. Otherwise there's too much impedance mismatch with the current implementation.

nucleus/deploy/nucleus_integration.py
```diff
@@ -0,0 +1,156 @@
+# This file contains all the Core Nucleus <-> HMI integrations
```
yixu34, line 1: @sasha-scale @ardila would be good to get your 👀 on this. Bear in mind this is temporary - the orchestration code in here would likely be moved to some Nucleus API endpoint (as discussed today, regarding https://app.shortcut.com/scaleai/story/350521/amend-model-creation-api-endpoint-to-optionally-be-associated-with-a-model-bundle-id-if-it-exists-write-to-nucleus-model)

yixu34: cc @Tianwei-She

nucleus/deploy/find_packages.py
yixu34: Curious, what changes?

seanshi-scale: mostly lint changes, I also changed a line to `if module_name == "nucleus"` somewhere, and had some functions at the end return values instead of just printing them out

nucleus/deploy/client.py
```diff
@@ -0,0 +1,271 @@
...
+
+        self.connection.post(
+            payload=dict(id=model_bundle_name, location=raw_s3_url),
+            route="model_bundle",
+        )  # TODO use return value somehow
+        # resp["data"]["bundle_name"] should equal model_bundle_name
+        # TODO check that a model bundle was created and no name collisions happened
+        return ModelBundle(model_bundle_name)
+
+    def create_model_endpoint(
```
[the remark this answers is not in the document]
seanshi-scale: I'm not super sure if we currently have a distinction between "sync" and "async" endpoints, i.e. right now everything is celery which in my mind means "async". Our synchronous request endpoints just wait for celery tasks to complete.

nucleus/deploy/client.py
```diff
@@ -0,0 +1,271 @@
...
+
+        """
+        if requirements is None:
+            requirements_inferred = find_packages_from_imports(globals())
+            requirements = [
+                f"{key}=={value}"
+                for key, value in requirements_inferred.items()
+            ]
+            logger.info(
+                "Using \n%s\n for model endpoint %s",
```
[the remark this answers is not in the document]
seanshi-scale: `pylint` told me to not use f-strings lol

nucleus/deploy/nucleus_integration.py
syandroo: Hmm this stuff does add the model upload + bundle storing in the db in one go, might wanna do something similar in the Nucleus API endpoint we talked about

nucleus/deploy/nucleus_integration.py
```diff
@@ -0,0 +1,156 @@
...
+    dataset: Dataset,
+) -> Tuple[List[str], Dict[str, DatasetItem]]:
+    # TODO I'm not sure if dataset items are necessarily s3URLs. Does this matter?
+    # TODO support lidar point clouds
+    if len(dataset.items) == 0:
+        logger.warning("Passed a dataset of length 0")
+        return [], {}  # TODO return type?
+    dataset_item_type = dataset.items[0].type
+    if not all(data.type == dataset_item_type for data in dataset.items):
+        logger.warning("Dataset has multiple item types")
```
ardila, line 121: This is going to be normal for scene datasets that have a mix of pointcloud and image data. Our approach on the autotag/model zoo side has been to use only the image data. But you may want to think of this more carefully in case people want to host models that are 3D.

nucleus/deploy/nucleus_integration.py
```diff
@@ -0,0 +1,156 @@
...
+        raise Exception  # TODO (code style) too broad exception
+
+    s3url_to_dataset_map = {}
+    # Do we need to keep track of nucleus ids?
+    if dataset_item_type == DatasetItemType.IMAGE:
+        s3Urls = [
+            data.image_location
+            for data in dataset.items
+            if data.image_location is not None
```
ardila, line 130: This is going to be the customer location of the URL. If they are doing hosted model inference, I believe we want to run on the processed_url, which isn't currently exported as part of the dataset item.

ardila (approved):
Approving since this isn't master but I think you need to have a clearer story around how HMI will work on Nucleus datasets

1. That have multiple types (common)
2. Where the customer wants to run HMI on our side, so they should use the images that are on our side, and not the original image URLs.

nucleus/deploy/client.py
```diff
@@ -0,0 +1,286 @@
...
+            )
+        serialized_bundle = cloudpickle.dumps(bundle)
+        requests.put(s3_path, data=serialized_bundle)
+
+        self.connection.post(
+            payload=dict(id=model_bundle_name, location=raw_s3_url),
+            route="model_bundle",
+        )  # TODO use return value somehow
+        # resp["data"]["bundle_name"] should equal model_bundle_name
+        # TODO check that a model bundle was created and no name collisions happened
```
sasha-scale, line 91: The connection class already raises an error for a non-200 status code: https://github.com/scaleapi/nucleus-python-client/blob/b18c412367e2f8ff8a7ba24e1cdb25b11b204b1b/nucleus/connection.py#L73

I would advise returning non-200 on name collision

nucleus/deploy/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+Moving an ML model from experiment to production requires significant engineering lift.
+Scale Deploy provides ML engineers a simple Python interface for turning a local code snippet into a production service.
+A ML engineer simply needs to call a few functions from Scale's SDK, which quickly spins up a production-ready service.
+The service efficiently utilizes compute resources and automatically scales according to traffic.
+
+
+Central to Scale Deploy are the notions of a `ModelBundle` and a `ModelEndpoint`.
+
+A `ModelBundle` consists of a trained model as well as the surrounding preprocessing and postprocessing code.
+Specifically, a `ModelBundle` consists of two Python objects, a (`model` or `load_model`), and a `load_predict_fn`; such that
```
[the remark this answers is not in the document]
seanshi-scale: It is expected to have exactly one of `model` or `load_model_fn`

nucleus/deploy/client.py
```diff
@@ -0,0 +1,286 @@
...
+            )
+
+        # Grab a signed url to make upload to
+        model_bundle_s3_url = self.connection.post({}, "model_bundle_upload")
+        if "signedUrl" not in model_bundle_s3_url:
+            raise Exception(
+                "Error in server request, no signedURL found"
+            )  # TODO code style broad exception
+        s3_path = model_bundle_s3_url["signedUrl"]
+        raw_s3_url = f"s3://{model_bundle_s3_url['bucket']}/{model_bundle_s3_url['key']}"
```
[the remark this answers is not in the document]
seanshi-scale: `response['signedUrl']` needs to be a signed URL as opposed to a raw s3 path, to ensure that the client is able to upload the model bundle to s3 at that location.

nucleus/deploy/client.py
```diff
@@ -0,0 +1,286 @@
...
+            model: Typically a trained Neural Network, e.g. a Pytorch module
+            load_model_fn: Function that when run, loads a model, e.g. a Pytorch module
+            load_predict_fn: Function that when called with model, returns a function that carries out inference
+        """
+
+        if (model is not None and load_model_fn is not None) or (
+            model is None and load_model_fn is None
+        ):
+            raise ValueError(
+                "Exactly one of model and load_model_fn should be non-None"
```
Tianwei-She, line 64: nit: add this requirement in docstring

nucleus/deploy/client.py
```diff
@@ -0,0 +1,271 @@
...
+        load_predict_fn(model) returns a function predict_fn that takes in model input and returns model output.
+        Pre/post-processing code can be included inside load_predict_fn/model.
+
+        Parameters:
+            model_bundle_name: Name of model bundle you want to create. This acts as a unique identifier.
+            model: Typically a trained Neural Network, e.g. a Pytorch module
+            load_predict_fn: Function that when called with model, returns a function that carries out inference
+        """
+        # Grab a signed url to make upload to
+        model_bundle_s3_url = self.connection.post({}, "model_bundle_upload")
```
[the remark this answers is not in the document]
seanshi-scale: The endpoint is defined at `server/routes/v1/hosted_inference.ts`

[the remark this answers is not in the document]
seanshi-scale: Apparently the `self.connection.post` blows up on a non-200 as well, so I think we can safely assume the request has returned 200

nucleus/deploy/nucleus_integration.py
seanshi-scale: Good point, we'll need to flesh out how Nucleus interacts with HMI a bit more I think. Possibly one thing we could do is have users define if an HMI ModelBundle/ModelEndpoint does 2D, 3D, or general inference, where if a ModelBundle is 2D or 3D, then it needs to conform to a pretty strict API. Otherwise, we'd somehow have to automatically translate between data types (e.g. detect when a bbox is (x, y, w, h) or (x1, y1, x2, y2)) which sounds close to impossible.

nucleus/deploy/nucleus_integration.py
seanshi-scale: Yeah this is true. I think the longer term plan is to have Nucleus itself call HMI, which makes this file unnecessary once that's built out

nucleus/deploy/nucleus_integration.py
yixu34: I think both the bundle and endpoint would be a black box, i.e. it's up to the user to define if it does 2D, 3D, or mixed inference, without needing explicit structured tagging from them. They'd also be responsible for providing the preprocessing/postprocessing code to translate into the necessary Nucleus python client types, which should be doable since the requests in this context are originating from the Nucleus client.

nucleus/deploy/nucleus_integration.py
yixu34: Ok so I guess we need to do some work on our (Scale's) side to expose this field? If so, I can create a ticket for the ML infra team to tackle this (unless Nucleus was planning on doing this anyway).

nucleus/deploy/client.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def async_request(self, endpoint_id: str, s3url: str) -> str:
+        """
+        Makes a request to the Model Endpoint at endpoint_id, and immediately returns a key that can be used to retrieve
+        the result of inference at a later time.
+
+        Parameters:
+            endpoint_id: The id of the endpoint to make the request to
+            s3url: A url that points to a file containing model input.
+                Must be accessible by Scale Deploy, hence it needs to either be public or a signedURL.
```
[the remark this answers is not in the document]
yixu34: +1 I would suggest that we go a bit more abstract and expose everything as URIs so as to be cloud agnostic.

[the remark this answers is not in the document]
seanshi-scale: Any public url should do, the format requirement is "openable by smart_open". Added a note on formats in model_endpoint.py

nucleus/deploy/client.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def list_bundles(self) -> List[ModelBundle]:
+        """
+        Returns a list of model bundles that the user owns.
+        TODO this route doesn't exist serverside
+        """
+        # resp = self.connection.get("model_bundle")
+        raise NotImplementedError
+
+    def list_model_endpoints(self) -> List[AsyncModelEndpoint]:
```
Tianwei-She, line 196: nit - use consistent name: model_endpoint vs endpoint, model_bundle vs bundle

perhaps we should just delete all "model" since it's kind of meaningless?

yixu34 (approved):
Once we get the pyspark <-> HMI integration working, I think it's OK to merge this. We can take another pass and address the outstanding comments and go over things in more detail.

PR #205, opened 2022-01-21
Upload custom evaluation functions

phil-scale:
Adds interface to upload custom evaluation function [sc-354699](https://app.shortcut.com/scaleai/story/354699)

Also improves speed of polygon intersection algorithm.

Python notebook in branch `phil/custom_metrics_notebook`. Link: https://github.com/scaleapi/nucleus-python-client/blob/phil/custom_metrics_notebook/demo.ipynb

### DIFF
--- a/nucleus/metrics/geometry.py
+++ b/nucleus/metrics/geometry.py
@@ -146,20 +146,20 @@ def area(box):
     return (box[2] - box[0]) * (box[3] - box[1])
 
 
-def iou(boxA, boxB):
-    boxC = intersection(boxA, boxB)
-    return area(boxC) / (area(boxA) + area(boxB) - area(boxC))
+def iou(box_a, box_b):
+    box_c = intersection(box_a, box_b)
+    return area(box_c) / (area(box_a) + area(box_b) - area(box_c))
 
 
-def intersection(boxA, boxB):
+def intersection(box_a, box_b):
     """boxes are left, top, right, bottom where left < right and top < bottom"""
-    boxC = [
-        max(boxA[0], boxB[0]),
-        max(boxA[1], boxB[1]),
-        min(boxA[2], boxB[2]),
-        min(boxA[3], boxB[3]),
+    box_c = [
+        max(box_a[0], box_b[0]),
+        max(box_a[1], box_b[1]),
+        min(box_a[2], box_b[2]),
+        min(box_a[3], box_b[3]),
     ]
-    return boxC
+    return box_c
 
 
 def rectangle_intersection_area(
--- a/nucleus/metrics/geometry.py
+++ b/nucleus/metrics/geometry.py
@@ -5,48 +5,27 @@ import numpy as np
 TOLERANCE = 1e-8
 
 
-class GeometryPoint:
-    def __init__(self, xy: Union[Tuple[float, float], np.ndarray]):
-        self.xy = np.array(xy)
-        self.x = xy[0]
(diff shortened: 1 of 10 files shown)


nucleus/modelci/client.py
```diff
@@ -29,9 +37,54 @@ def __repr__(self):
...
+        ), f"Expected eval_function to be of type Metric but got {type(eval_function)}"
+        response = self.connection.post(
+            EvalFunctionInput.from_metric(eval_function, name).dict(),
+            "modelci/eval_fn",
+        )
+        print(replace_double_slashes(response[LOGS_KEY]))
+        if response["eval_fn"] is None:
+            raise UploadEvalFunctionError(
+                "Eval function upload failed. See logs for traceback."
+            )
```
gatli, line 80: I'd create a `pydantic` model for this. I'd actually use a `pydantic` model for everything that goes over the wire since it helps with catching common formatting errors/ unexpected values.

You could actually replace the double slashes in the `pydantic` model and you could move this validation logic to a `@validator`

nucleus/modelci/data_transfer_objects/eval_function.py
```diff
@@ -79,3 +84,36 @@ class GetEvalFunctions(ImmutableModel):
...
+        )
+
+    @validator("serialized_fn")
+    def valid_serialized_fn(cls, v):  # pylint: disable=no-self-argument
+        deserialized = cloudpickle.loads(base64.b64decode(v))
+        if not isinstance(deserialized, Metric):
+            raise ValueError(
+                f"Expected field to be an instance of Metric but got {deserialized}"
+            )
+        return v
```
gatli, line 119: ❤️ nice use of `pydantic`

gatli (approved):
LGTM 👍

pfmark (approved):
Great work overall! 

I tested it and the setup and computation seem to work, also tested in the FE. 

I just ran into this bug during the upload:
Screen Shot 2022-01-25 at 20 26 37

And I couldn't delete eval functions. But I was running it against master and I guess should be running against `scaleapi/phil/model_ci_eval_fn_upload`? 

Let's address the comments, make sure the checks pass and figure out why my stuff broke, then let's merge!

nucleus/modelci/client.py
phil-scale: Yup, loving the validation and formatting that pydantic brings!

nucleus/metrics/geometry.py
```diff
@@ -157,25 +133,65 @@ def unique(ar):
...
+
+def intersection(boxA, boxB):
+    """boxes are left, top, right, bottom where left < right and top < bottom"""
+    boxC = [
+        max(boxA[0], boxB[0]),
+        max(boxA[1], boxB[1]),
+        min(boxA[2], boxB[2]),
+        min(boxA[3], boxB[3]),
+    ]
+    return boxC
```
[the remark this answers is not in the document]
phil-scale: ah the woes of copy-paste

pfmark (changes requested):
I tested it once more with the right backend. All in all still great. 

One thing that we should address though is that currently the user cannot delete a metric which has existing evaluations because of the foreign key constraint. 


````[


api] info: Successfully refreshed config cache {"label":"default"}
[api] info: [ModelCi] entered handleSendRequest with eval function ID ef_c7rq9me8sqfk8yvz7d70 anWe can keep the behavior as is for now

PR #209, opened 2022-02-01
Add support for easier metadata management 

jean-lucas:
Corresponding [backend PR](https://github.com/scaleapi/scaleapi/pull/35489)

Usage:

For now, please run it in the scale branch `jean-lucas-metadata-update` and start the backend.

```python

# get a dataset (ideally one with scenes), but any will do

mm = dataset.metadata_manager()

# Scene Item Update
scenes_items = mm.load_scenes('scene_1', with_items=True)  # or a list of ref ids
for item in scene_items:
    item.metadata['my_new_key'] = '123'
    for ds_item in item.dataset_items:
        ds_item.metadata['my_new_dataset_item_key'] = '456'

mm.update(scene_items)


# Dataset Item update
ds_items = mm.load_dataset_items('some_ref_id') # or a list of ref ids
for item in ds_items:
    item.metadata['hello'] = 'world'
mm.update(ds_items)
```

You should now be able to:
1. See this info in the sidebar of the detail view
2. Query on these new metadata keys

**TODOs in Future Iterations**:
- support key deletion
- support loading by modality, sensor name, or some other filtering criteria
- allow async updates for large jobs


drakejwong:
Sorry, don’t want you to have to rip out a bunch of code, but would it be easier to just return the `LidarScene` and `DatasetItem` objects themselves rather than load these new metadata objects? It seems the only additional attribute of the native objects is just the URL. This way we can also just leverage the existing `.append(update=True)` API rather than introducing a new util class to learn.

jean-lucas:
> but would it be easier to just return the LidarScene and DatasetItem objects themselves rather than load these new metadata objects?

I thought about it, but then would we leave all other fields empty? Or do we pull all of them in as well?
But I think it would make more sense to re-use these objects.

> This way we can also just leverage the existing `.append(update=True)` API rather than introducing a new util class to learn.

Not sure, because I cannot guarantee that the returning list from the database will be sorted in the same way as it was inserted. (A problem we faced last week).


I agree overall that this breaks consistency a bit, but my intentions was to keep this as lightweight as possible. The user has to only worry about reference_id and metadata, and ignore everything else.
I'll take another look at it, to see if I can incorporate some of your ideas.

PR #211, opened 2022-02-02
Remove body on GET requests for ScaleGov

edwardpark97:
Our ScaleGov environment has an AWS WAF (Web Application Firewall) that errors out on any GET request that has a json parameter (for some reason or another). GET requests really shouldn't have a body anyway so it shouldn't matter.


phil-scale (approved):

gatli (approved):
LGTM 👍

PR #214, opened 2022-02-07
Better COLAB installation

gatli:
COLAB has some pre-installed packages that cause problems when you update. You have to frequently restart the runtime and that makes for a very broken experience.

This allows click and scipy to be of the pre-installed version ... this gets around the problems mentioned. We don't use any specific click 8.0> functionality for now and have limited exposure on the scipy front. The used method has been there since 0.17.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ exclude = '''
 
 [tool.poetry]
 name = "scale-nucleus"
-version = "0.6.0"
+version = "0.6.1"
 description = "The official Python client library for Nucleus, the Data Platform for AI"
 license =  "MIT"
 authors = ["Scale AI Nucleus Team <nucleusapi@scaleapi.com>"]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -38,12 +38,10 @@ tqdm = "^4.41.0"
 dataclasses = { version = "^0.7", python = "^3.6.1, <3.7" }
 aiohttp = "^3.7.4"
 nest-asyncio = "^1.5.1"
-Sphinx = "^4.2.0"
```
gatli, line 41: Already in dev depends.

pyproject.toml
```diff
@@ -38,12 +38,10 @@ tqdm = "^4.41.0"
...
 pydantic = "^1.8.2"
-isort = "^5.10.1"
```
gatli, line 43: Purely a dev dependency.

pfmark:
Does this cause any problems when someone it running this outside of Colab? E.g. using Jupyter or just using the SDK itself?

gatli:
> Does this cause any problems when someone it running this outside of Colab? E.g. using Jupyter or just using the SDK itself?

I don't foresee any problems. In general it should choose to update to a later version but it won't throw a `contextualversionerror` in your face. This on the other hand fixes the COLAB issues. Win-win 🏅 🥇 🏆

pfmark:
okay lgtm then! Is the test a flaky test and can we get this to pass? Just making sure that it's nothing that got introduced here.

pfmark (approved):

gatli:
@pfmark It was a flake 😫

PR #216, opened 2022-02-10
Add basic metadata update for scenes/items

jean-lucas:
Provides a bare-bones approach to updating scenes _or_ dataset items metadata.

**Current approach user can take**

```python

print(dataset.scenes)  # get all top-level scene info, such as the `reference_id`
scene = dataset.get_scene('scene_1')

# add new keys to the existing ITEM metadata
for f in scene.frames:
    if 'lidar' in f.items.keys():
        f.items['lidar'].metadata['some_lidar_value'] = 123

# update the SCENE metadata
update_scene_meta = {**scene.metadata, **{'scene_foo': 123}}

# create the update scene object with its updated frames
updated_scene = nucleus.LidarScene(
    frames=scene.frames,
    reference_id='scene_1',  # this is DANGEROUS, if the user mistypes the reference id here. There is no check.
    metadata=update_scene_meta,
)

# commit the changes
dataset.append(items=[updated_scene], update=True, asynchronous=True)
```

**New approach with this PR**

```python
scene = dataset.get_scene('scene_1')

# Update Items metadata

lidar_ref_ids = [item['lidar'].ref_id for item in scene.frames if 'lidar' in item.keys()]
items_mapping = {}
for lidar_ref_id in lidar_ref_ids:
    items_mapping[lidar_ref_id] = {'some_lidar_value': 123}

dataset.update_items_metadata(items_mapping)

# Update scene metadata

dataset.update_scenes_metadata({'scene_1': {'scene_foo': 123}})
```

**Note**:
- This does not do any validation on deleted keys or invalid reference IDs. Its a simpler `MetadataManager` than before. Its up to the user to generate the correct mappings.


nucleus/dataset.py
```diff
@@ -1392,3 +1393,45 @@ def _upload_items(
 
         populator = DatasetItemUploader(self.id, self._client)
         return populator.upload(dataset_items, batch_size, update)
+
+    def update_scenes_metadata(self, mapping: Dict[str, dict]):
```
jean-lucas, line 1397: Up for discussion, but we can name these `update_scenes_metadata` to `update_scenes_metadata_raw`,
to communicate that little validation is done.

jean-lucas: Alternatively we can swap the parameters as such:
- `update_scenes_metadata(mapping: dict)` -> `update_scenes_metadata(List[Scene])`
- `update_items_metadata(mapping: dict)` -> `update_items_metadata(List[Frame | DatasetItem])`

Then just extract the mapping internally.

jean-lucas: I'm applying these two changes right now actually. Commit coming in 1-2 hours.

jean-lucas: Haven't started yet. But to clarify. You prefer to keep this as is?

jean-lucas: Cool, left it as is then.

PR #217, opened 2022-02-10
Speed up geometry functions

phil-scale:
Speeds up geometry functions for calculating polygon intersection, especially for rectangles.

### DIFF
--- a/nucleus/metrics/geometry.py
+++ b/nucleus/metrics/geometry.py
@@ -187,10 +187,8 @@ def polygon_intersection_area(
     nb = len(polygon_b)
     res = 0.0
     for i in range(1, na - 1):
-        # sa = [polygon_a[0], polygon_a[i], polygon_a[i + 1]]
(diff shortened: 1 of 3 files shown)


gatli (commented):
Nice 👍 

Did you happen to profile the runtime of this solution? I'm wondering where we're spending the most time.

phil-scale:
> Nice 👍
> 
> Did you happen to profile the runtime of this solution? I'm wondering where we're spending the most time.

Yes, I used cProfile and found that the majority of the time was being spent on `numpy` array creation. I think at some point I might convert this to a `numba` program so it compiles jit.

gatli:
> > Nice 👍
> > Did you happen to profile the runtime of this solution? I'm wondering where we're spending the most time.
> 
> Yes, I used cProfile and found that the majority of the time was being spent on `numpy` array creation. I think at some point I might convert this to a `numba` program so it compiles jit.

Fun! LMK if you start trying it out 🙂

PR #218, opened 2022-02-10
Internal deploy python client functionality

seanshi-scale:
The client operates in two modes, external mode (default) and self-hosted mode.

Also add a few extra functions for the missing API calls (e.g. delete bundle/endpoint, put endpoint)

For uploading model bundles in self-hosted mode, there is an `upload_bundle_fn` that needs to be passed to the client. This allows clients to specify their own model bundle storage backends. An example implementation is
```
def upload_bundle_fn(serialized_bundle, bundle_url):
    transport_params = {"session": Session(profile_name="ml-worker")}
    with smart_open.open(bundle_url, "wb", transport_params=transport_params) as f:
        f.write(serialized_bundle)
```

n.b. the code assumes that the REST APIs for the public route and the internal route are nearly identical, e.g. creating a model bundle is always at `POST /model_bundle/` for example, and it takes in a JSON with a key `url`.

Testing:

Ran a testing script that
- creates a model bundle
- creates a model endpoint
- sends a task through
- deletes the endpoint
- deletes the bundle

using the client in internal and external mode. Calls all succeed, the task doesn't return a result because of some other bug though.


nucleus/deploy/client.py
```diff
@@ -28,28 +28,53 @@
...
+        self.is_internal = is_internal
+        self.upload_bundle_fn: Optional[Callable[[str, str], None]] = None
 
     def __repr__(self):
         return f"DeployClient(connection='{self.connection}')"
 
     def __eq__(self, other):
         return self.connection == other.connection
 
+    def register_upload_bundle_fn(
```
seanshi-scale, line 55: Not sure if this is the best way of allowing users to use their own backend for storing model bundles (in internal mode).

nucleus/deploy/client.py
```diff
@@ -28,28 +28,53 @@
...
         Initializes a Scale Deploy Client.
 
         Parameters:
             api_key: Your Scale API key
             endpoint: The Scale Deploy Endpoint (this should not need to be changed)
+            is_internal: True iff you are connecting to a self-hosted Scale Deploy
         """
         self.connection = Connection(api_key, endpoint)
+        self.is_internal = is_internal
+        self.upload_bundle_fn: Optional[Callable[[str, str], None]] = None
```
Tianwei-She, line 47: line 47 should be deleted?

nucleus/deploy/client.py
```diff
@@ -170,11 +215,20 @@ def create_model_endpoint(
             per_worker=per_worker,
             requirements=requirements,
         )
+        if self.is_internal:
+            payload.update(
+                aws_role=aws_role, results_s3_bucket=results_s3_bucket
```
Tianwei-She, line 220: should check `aws_role` and `results_s3_bucket` are not None

nucleus/deploy/client.py
seanshi-scale: It gets set elsewhere, but I think people (and my IDE) say it's good to initialize all instance variables inside __init__

Tianwei-She (approved):
I'm not sure if this is the best design, but I think it's ok to merge for now

nucleus/deploy/client.py
```diff
@@ -28,28 +28,53 @@
 class DeployClient:
     """Scale Deploy Python Client extension."""
 
-    def __init__(self, api_key: str, endpoint: str = SCALE_DEPLOY_ENDPOINT):
+    def __init__(
+        self,
+        api_key: str,
+        endpoint: str = SCALE_DEPLOY_ENDPOINT,
```
syandroo, line 34: hmm should we have this as a default? won't this not work if it's wrong

nucleus/deploy/client.py
```diff
@@ -28,28 +28,53 @@
...
+        is_internal: bool = False,
```
syandroo, line 35: hmm didn't we say we weren't gonna have an internal / external split so we don't commit this to a public repo?

ok with it i guess

nucleus/deploy/client.py
```diff
@@ -113,9 +150,13 @@ def create_model_endpoint(
         env_params: Dict[str, str],
         requirements: Optional[List[str]] = None,
         gpu_type: Optional[str] = None,
+        aws_role: Optional[str] = None,
```
yixu34, line 153: I think this is leaking too many implementation details. Showing people that we use s3 for signed URL uploads should be fine because they're signed (though maybe we still want to think long term about how we accept data).

Maybe this was already a security hole and this change is only highlighting the fact, but doesn't this mean that I could modify the client library and pass in my AWS role of choice to Scale's hosted Deploy service?

I would suggest going a bit more abstract and following a DI (dependency injection)-esque approach. You can pass in an optional "AuthDecorator" object/function, where external users pass `None` (the default), and internal users pass in something that lives in our internal repo that enriches this payload with the role info. This way, you don't expose as many details about AWS, which makes our cleanup later easier. Even this isn't the best abstraction, but it'd still be better IMO.

But we'd still need to address the underlying security issue. I would suggest that we also do server-side validation, where we strip out the `aws_role` and `results_s3_bucket` in the payload in the Gateway service only if a scaleapi-auth'ed `userid` is present (i.e. it's an external user). @syandroo thoughts?

yixu34: Discussed offline - I misunderstood and forgot that we are forcing the right `aws_role` in `scaleapi` server-side.

yixu34 (approved):

PR #219, opened 2022-02-14
Delete datasets after every creation

gatli:
There were a couple of places where datasets weren't being cleaned up after test runs. This makes sure we aren't littering.

### DIFF
--- a/tests/cli/test_jobs.py
+++ b/tests/cli/test_jobs.py
@@ -19,3 +19,4 @@ def test_list_jobs(CLIENT, runner):
     time.sleep(0.5)
     assert result.exception is None
     assert result.exit_code == 0
+    CLIENT.delete_dataset(dataset.id)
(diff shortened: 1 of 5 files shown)


tests/cli/test_jobs.py
```diff
@@ -19,3 +19,4 @@ def test_list_jobs(CLIENT, runner):
     time.sleep(0.5)
     assert result.exception is None
     assert result.exit_code == 0
+    CLIENT.delete_dataset(dataset.id)
```
hendrikmakait, line 22: Can we refactor this test to use a fixture or setup/teardown? When an assertion fails, we won't delete the dataset otherwise.

hendrikmakait (approved):
LGTM overall, one suggestion to refactor creation and deletion into setup/teardown functionality to avoid littering on test failures.

tests/cli/test_jobs.py
gatli: Yes, of course. Lazy monday thinking 🙂

PR #224, opened 2022-02-22
Add LineAnnotation and LinePrediction support

flubstep:


nucleus/annotation.py
```diff
@@ -46,18 +47,17 @@ class Annotation:
...
-            return PolygonAnnotation.from_json(payload)
-        elif payload.get(TYPE_KEY, None) == CUBOID_TYPE:
-            return CuboidAnnotation.from_json(payload)
-        elif payload.get(TYPE_KEY, None) == CATEGORY_TYPE:
-            return CategoryAnnotation.from_json(payload)
-        elif payload.get(TYPE_KEY, None) == MULTICATEGORY_TYPE:
-            return MultiCategoryAnnotation.from_json(payload)
-        else:
-            return SegmentationAnnotation.from_json(payload)
+        type_key_to_type: Dict[str, Type[Annotation]] = {
```
flubstep, line 50: Needed to move this to a dict lookup because `pylint` will block if you have more than 6 `return`s in one function.

drakejwong (approved):

PR #225, opened 2022-02-22
Add making requests to synchronous endpoints

seanshi-scale:
Companion pr to https://github.com/scaleapi/models/pull/2584

Work for both https://app.shortcut.com/scaleai/story/381256/ and https://app.shortcut.com/scaleai/story/379721/


nucleus/deploy/client.py
```diff
@@ -263,24 +274,31 @@ def list_model_endpoints(self) -> List[AsyncModelEndpoint]:
...
+            AsyncModelEndpoint(endpoint_id=endpoint["name"], client=self)
+            for endpoint in resp["endpoints"]
+            if endpoint["endpoint_type"] == "async"
+        ]
+        sync_endpoints: List[Union[AsyncModelEndpoint, SyncModelEndpoint]] = [
+            SyncModelEndpoint(endpoint_id=endpoint["name"], client=self)
+            for endpoint in resp["endpoints"]
+            if endpoint["endpoint_type"] == "sync"
         ]
+        return async_endpoints + sync_endpoints
```
yixu34, line 287: Don't the set of async and sync endpoints comprise the full set of possible endpoints? If so, can we just return `resp["endpoints"]`?

nucleus/deploy/model_endpoint.py
```diff
@@ -7,6 +7,22 @@
...
 
+class SyncModelEndpoint:
+    def __init__(self, endpoint_id: str, client):
+        self.endpoint_id = endpoint_id
+        self.client = client
+
+    def __str__(self):
+        return f"SyncModelEndpoint <endpoint_id:{self.endpoint_id}>"
+
+    def predict_one(self, url):
```
yixu34, line 18: Can we just call this `predict`?

Also, non-url inputs are in a different ticket, right?

yixu34 (approved):

ivan-khomyakov-sc (approved):

nucleus/deploy/client.py
seanshi-scale: It is the full set of endpoints, but the object returned is a list of either `SyncModelEndpoints` or `AsyncModelEndpoints`, so we have to construct them separately (or do something like `modelEndpointClass = SyncModelEndpoint if <x> else Async`)

nucleus/deploy/model_endpoint.py
seanshi-scale: re non-url inputs, they are in a different ticket, and I'm inclined to just open a different pr for them as well at this point

re `predict`, have offline discussion

PR #229, opened 2022-02-25
[Validate] Pass Evaluation Function arguments with EvaluationCriteria

gatli:
This adds the possibility of configuring eval functions with keyword arguments. 

I've deliberately not added many checks on the client side since I think that should be handled by the server since that allows us to maintain compatibility with a wider range of SDK versions.


nucleus/validate/client.py
```diff
@@ -78,6 +83,30 @@ def create_scenario_test(
                 "Must pass an evaluation_criteria to the scenario test! I.e. "
                 "evaluation_criteria = [client.validate.eval_functions.bbox_iou() > 0.5]"
             )
+        incorrect_type = [
```
gatli, line 86: More concrete error message for (the probably common case) of forgetting to add a comparison operator to the `evaluation_criteria`

nucleus/validate/client.py
```diff
@@ -11,12 +11,17 @@
...
 from .eval_functions.available_eval_functions import AvailableEvalFunctions
+from .eval_functions.base_eval_function import BaseEvalFunction
 from .scenario_test import ScenarioTest
 
 SUCCESS_KEY = "success"
 EVAL_FUNCTIONS_KEY = "eval_functions"
 
 
+class InvalidEvaluationCriteria(Exception):
+    pass
```
phil-scale, line 22: nit: I think this should go in the `errors.py` file

nucleus/validate/client.py
```diff
@@ -78,6 +83,30 @@ def create_scenario_test(
                 "Must pass an evaluation_criteria to the scenario test! I.e. "
                 "evaluation_criteria = [client.validate.eval_functions.bbox_iou() > 0.5]"
             )
+        incorrect_type = [
+            crit
+            for crit in evaluation_criteria
+            if not isinstance(crit, EvaluationCriterion)
+        ]
```
phil-scale, line 90: Might be more readable using `all`
```suggestion
        incorrect_type = all([isinstance(crit, EvaluationCriterion) for crit in evaluation_criteria])
```

phil-scale (approved):
Overall looks pretty reasonable to me - a couple of minor nits

nucleus/validate/eval_functions/available_eval_functions.py
```diff
@@ -11,30 +11,147 @@
...
     @classmethod
     def expected_name(cls) -> str:
         return "bbox_map"
 
 
 class BoundingBoxRecall(BaseEvalFunction):
+    def __call__(
+        self,
+        enforce_label_match: bool = False,
+        iou_threshold: float = 0.0,
```
pfmark, line 65: Isn't the threshold for detection matching under the hood 0.5? Then we should go for default 0.5 here too, otherwise this doesn't make sense

nucleus/validate/eval_functions/available_eval_functions.py
```diff
@@ -11,30 +11,147 @@
...
+            **kwargs,
+        )
+
     @classmethod
     def expected_name(cls) -> str:
         return "bbox_recall"
 
 
 class BoundingBoxPrecision(BaseEvalFunction):
+    def __call__(
```
pfmark, line 89: Let's also add a usage example to the docs of each of those evaluation functions

pfmark (approved):

nucleus/validate/client.py
gatli: Good nit 🙂

nucleus/validate/client.py
gatli: I'm actually using implicit list length check, an empty list evaluates to `False` and then using the `incorrect_type` as a list for creating a better error message.

nucleus/validate/eval_functions/available_eval_functions.py
gatli: Not sure, the default parameters are copied from the `nucleus/metrics` module. I'll check!

nucleus/validate/eval_functions/available_eval_functions.py
gatli: Good idea!

nucleus/validate/eval_functions/available_eval_functions.py
gatli: Yeah, copy paste error on my side. Good catch 🙂

nucleus/validate/client.py
gatli: Renamed to `incorrect_types` and made a length check to make it look more list-y.

sasha-scale (approved):

PR #231, opened 2022-02-26
send code of the function to the endpoint

ivan-khomyakov-sc:
Upload source code with the bundle. It will be stored in metadata field


seanshi-scale (approved):
lgtm!

syandroo:
but @ivan-khomyakov-sc we should have a follow up to this to move it to its own column in the db, i'll slack you when i write the readme for migrations

PR #232, opened 2022-02-26
Python client changes for json inputs

seanshi-scale:
Python client handles json/serialized inputs + outputs, also add a `bundle_location_fn` notion.

Added an `EndpointRequest` class to encapsulate the options for a request to an endpoint

### DIFF
--- a/nucleus/deploy/model_endpoint.py
+++ b/nucleus/deploy/model_endpoint.py
@@ -1,6 +1,6 @@
 import concurrent.futures
 from collections import Counter
-from typing import Dict, Optional, Sequence
+from typing import Any, Dict, Optional, Sequence
 
 TASK_PENDING_STATE = "PENDING"
 TASK_SUCCESS_STATE = "SUCCESS"
@@ -161,13 +161,15 @@ class AsyncModelEndpointResponse:
     ):
 
         self.client = client
-        self.request_ids = request_ids.copy()  # url or str(args) -> task_id
-        self.responses: Dict[str, Optional[str]] = {
-            url: None for url in request_ids.keys()
+        self.request_ids = (
+            request_ids.copy()
+        )  # custom request_id or url or str(args) -> task_id
+        self.responses: Dict[str, Optional[Dict]] = {
+            req_id: None for req_id in request_ids.keys()
         }
         # celery task statuses
         self.statuses: Dict[str, Optional[str]] = {
-            url: TASK_PENDING_STATE for url in request_ids.keys()
+            req_id: TASK_PENDING_STATE for req_id in request_ids.keys()
         }
 
     def poll_endpoints(self):
@@ -177,15 +179,16 @@ class AsyncModelEndpointResponse:
 
         # TODO: replace with batch endpoint, or make requests in parallel
 
-        def single_request(inner_url, inner_request_id):
+        def single_request(inner_url, inner_task_id):
             if self.statuses[inner_url] != TASK_PENDING_STATE:
                 return None
-            inner_response = self.client.get_async_response(inner_request_id)
+            inner_response = self.client.get_async_response(inner_task_id)
+            print("inner response", inner_response)
             return (
                 inner_url,
-                inner_request_id,
+                inner_task_id,
                 inner_response.get("state", None),
-                inner_response.get("result_url", None),
+                inner_response,
             )
 
         with concurrent.futures.ThreadPoolExecutor(max_workers=20) as executor:
@@ -204,16 +207,6 @@ class AsyncModelEndpointResponse:
             if result_url:
                 self.responses[url] = result_url
 
-        # for url, request_id in self.request_ids.items():
-        #     current_state = self.statuses[url]
-        #     if current_state == TASK_PENDING_STATE:
-        #         response = self.client.get_async_response(request_id)
-        #         print(response)
-        #         if "state" in response:
-        #             self.statuses[url] = response["state"]
-        #         if "result_url" in response:
-        #             self.responses[url] = response["result_url"]
-
     def is_done(self, poll=True) -> bool:
         """
         Checks if all the tasks from this round of requests are done, according to
@@ -227,7 +220,7 @@ class AsyncModelEndpointResponse:
             resp != TASK_PENDING_STATE for resp in self.statuses.values()
         )
 
-    def get_responses(self) -> Dict[str, Optional[str]]:
+    def get_responses(self) -> Dict[str, Optional[Dict[Any, Any]]]:
         if not self.is_done(poll=False):
             raise ValueError("Not all responses are done")
         return self.responses.copy()
--- a/nucleus/deploy/model_endpoint.py
+++ b/nucleus/deploy/model_endpoint.py
@@ -31,6 +31,8 @@ class EndpointRequest:
         return_pickled: Optional[bool] = True,
         request_id: Optional[str] = None,
     ):
+        # TODO: request_id is pretty much here only to support the clientside AsyncModelEndpointBatchResponse
+        # so it should be removed when we get proper batch endpoints working.
         validate_task_request(url=url, args=args)
         if request_id is None:
             request_id = str(uuid.uuid4())
--- a/nucleus/deploy/model_endpoint.py
+++ b/nucleus/deploy/model_endpoint.py
@@ -19,7 +19,8 @@ class EndpointRequest:
             predict_fn(foo, bar), then the keys in the dictionary should be 'foo' and 'bar'. Values must be native Python
             objects.
         return_pickled: Whether the output should be a pickled python object, or directly returned serialized json
-        request_id: A user-specifiable id for requests. Should be unique.
+        request_id: A user-specifiable id for requests.
+            Should be unique among EndpointRequests made in the same batch call.
             If one isn't provided the client will generate its own.
     """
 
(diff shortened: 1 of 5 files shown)


nucleus/deploy/model_endpoint.py
```diff
@@ -57,28 +90,27 @@ def predict_batch(
...
                 endpoint_id=self.endpoint_id,
-                url=inner_url,
+                url=request.url,
+                args=request.args,
+                return_pickled=request.return_pickled,
             )
-            return inner_url, inner_inference_request
+            if request.request_id is not None:
+                request_key = request.request_id
+            elif request.url is not None:
```
seanshi-scale, line 104: maybe this stuff really isn't necessary :P

nucleus/deploy/model_endpoint.py
```diff
@@ -1,12 +1,41 @@
...
 from collections import Counter
-from typing import Dict, Optional, Sequence
+from typing import Any, Dict, Optional, Sequence
 
 TASK_PENDING_STATE = "PENDING"
 TASK_SUCCESS_STATE = "SUCCESS"
 TASK_FAILURE_STATE = "FAILURE"
 
 
+class EndpointRequest:
```
seanshi-scale, line 10: since we have a function to predict on multiple inputs, I added this so people could pass in a mix of urls/jsons as inputs. Not sure if that'll happen

ivan-khomyakov-sc (approved):

yixu34 (commented):
Don't merge yet still commenting

nucleus/deploy/model_endpoint.py
yixu34: Might be good to add an `EndpointResponse` class for symmetry?

nucleus/deploy/model_endpoint.py
```diff
@@ -15,8 +44,13 @@ def __init__(self, endpoint_id: str, client):
     def __str__(self):
         return f"SyncModelEndpoint <endpoint_id:{self.endpoint_id}>"
 
-    def predict(self, url):
-        return self.client.sync_request(self.endpoint_id, url)
+    def predict(self, request: EndpointRequest):
```
yixu34, line 47: Per above, `-> EndpointResponse` might be a handy result and annotation.

nucleus/deploy/model_endpoint.py
yixu34: wdym?

nucleus/deploy/model_endpoint.py
```diff
@@ -1,12 +1,41 @@
...
+class EndpointRequest:
+    """
+    Represents a single request to either a SyncModelEndpoint or AsyncModelEndpoint.
+    Parameters:
+        url: A url to some file that can be read in to a ModelBundle's predict function. Can be an image, raw text, etc.
+        args: A Dictionary with arguments to a ModelBundle's predict function. If the predict function has signature
+            predict_fn(foo, bar), then the keys in the dictionary should be 'foo' and 'bar'. Values must be native Python
+            objects.
+        return_pickled: Whether the output should be a pickled python object, or directly returned serialized json
+        request_id: A user-specifiable id for requests. Should be unique.
```
yixu34, line 19: When/why would a user specify this? I was imagining this would be populated by us server-side.

nucleus/deploy/model_endpoint.py
```diff
@@ -1,12 +1,41 @@
...
+    """
+
+    def __init__(
+        self,
+        url: Optional[str] = None,
+        args: Optional[Dict] = None,
```
yixu34, line 25: Wonder if we should have these two things be a `Union` (we'd still have to do the XOR check because `Union` is really just a hint, from a type annotation perspective).

nucleus/deploy/model_endpoint.py
```diff
@@ -1,12 +1,41 @@
...
+        return_pickled: Optional[bool] = True,
+        request_id: Optional[str] = None,
+    ):
+        if url is None and args is None:
```
yixu34, line 29: Related to the `Union` comment, think this is happening enough where we'd want a helper function.

nucleus/deploy/model_endpoint.py
seanshi-scale: The thing this `request_id` is trying to solve is that a user needs to be able to associate model inference results with a given model input. For example, for the nucleus integration we need to maintain some map from s3URL of input images to task results, so this request_id can be set to that s3URL. Before, I was implicitly using the url as the request_id, but we don't necessarily have urls anymore.

I suppose that we could just keep the association implicitly in the order of some list somewhere but then the user would have to do that too.

yixu34: Ok it sounds like you're basically using this as a `reference_id` then? Can we just have our client generate a UUID? Then the nucleus integration (which one day will be server-side) can just maintain that `Request` -> `Response` mapping, with the hash key being this UUID.

nucleus/deploy/model_endpoint.py
seanshi-scale: there's a lot of logic to try to figure out a reasonable request_id if the user didn't pass one in, e.g. use the url, or the args, which seems "a bit extra"

nucleus/deploy/model_endpoint.py
seanshi-scale: I sorta am leaning away from putting the two into a `Union`. We'd have to keep track of whether the data represents a url or args (which we technically could do by looking at their types, but this feels wrong), and keeping them separate seems like the easiest way to do this. The stuff stored at url/args gets put into separate fields in the eventual http request's json anyway, so I think it's simpler to keep things separate.

nucleus/deploy/model_endpoint.py
yixu34: Oh ok, as per my other comment ([18 hours ago](https://github.com/scaleapi/nucleus-python-client/pull/232#discussion_r818259959)), can we just generate a UUID?

nucleus/deploy/model_endpoint.py
seanshi-scale: I left in the ability for clients to specify their own `request_id`s, but otherwise the client will generate a uuid for the `request_id`

yixu34: I would err on the more restrictive side and not allow `request_id`s, because I'm imagining a hapless user passing in "42" hardcoded and never changing it.

nucleus/deploy/model_endpoint.py
yixu34: Yup that makes sense.

nucleus/deploy/nucleus_integration.py
```diff
@@ -26,7 +26,7 @@ class NucleusDatasetInferenceRun:
     # For the demo, we will need our Nucleus Dataset to have `image_location`s in s3://scale-ml-hosted-model-inference
     def __init__(
         self,
-        hmi_async_job: AsyncModelEndpointResponse,
+        hmi_async_job: AsyncModelEndpointBatchResponse,
```
yixu34, line 29: Not really part of this PR but maybe we can take a pass at removing all "hmi" stuff from user-facing names?

nucleus/deploy/model_endpoint.py
seanshi-scale: (probably should have specified this but) The uniqueness condition is actually more permissive than "globally unique", but instead is "the EndpointRequests producing a given BatchResponse must have distinct `request_ids`". I think that mostly gets rid of the hardcoding problem? (i.e. it doesn't matter if the user runs the script twice with the same value of `request_id`, but if the user hardcodes a value for `request_id` somewhere and copies the code that would be a problem)

I do think we should allow users to set `request_id`s though, since otherwise it's much more annoying for users to keep track of an association from `EndpointRequest`/`EndpointResponse` to something a user cares about, e.g. if the user wants to run inference on N images, they need to keep track of which `EndpointResponse` gets associated with which image. 

With user-specified `request_id`s the user can do `endpoint_response_for_image_i = job.responses[request_id_for_image_i]`; with non-user-specified `request_ids` the user would need to do 
```
# Go through the list of EndpointRequests to find the request_id corresponding to a given image, comparing either `url` or `args`. Comparing values seems reasonable in the `url` case but ugly in the `args` case.
# job.responses[the request_id we found]
```
or use the ordering of `EndpointRequest` fed in and the corresponding ordering of `EndpointResponses` (currently `EndpointResponses` are returned in a dictionary though)
(or something I haven't thought of)

One simple thing we could do to catch errors earlier is just to check that the user passes a set of distinct `request_id`s to the batch_predict function at the very least

yixu34: Oh I see. I think this is happening because the client library is still operating under the paradigm where batch jobs are tied to endpoints instead of being standalone. In the eventual state, I would expect the batch request to be its own class (not a `predict_batch` method off of an endpoint class), and result would be a url to a file/directory that contains outputs, with the relative order not being guaranteed anyway. In other words, this is needed only because have to support the current janky client-side integration. In that case I guess this is fine, but prior to launch I definitely would want `request_id` to be gone altogether from the client, so please add a TODO.

I was tempted to suggest that we call this `reference_id`, but that's misleading in a _different_ way (Nucleus actually persists those, so for us to use them as ephemeral client-side IDs could be confusing due to inconsistency).

seanshi-scale: Yeah it's pretty much there because of the async batch class

yixu34 (approved):

PR #237, opened 2022-03-03
Update hmi client to use new API

syandroo:


seanshi-scale (approved):
yeah this looks good lol

PR #238, opened 2022-03-03
Only check for partial equality

ardila:
* Since the backend may add new information, only check for some fields, don't expect perfect equality.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -546,3 +546,17 @@ def create_predictions(
     ]
     dataset.upload_predictions(model, predictions)
     return predictions
+
+
+def assert_partial_equality(expected, result, base_keys=[]):
+    """Used to check partial equality of two json-like dictionaries."""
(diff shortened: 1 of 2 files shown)


tests/helpers.py
```diff
@@ -546,3 +546,17 @@ def create_predictions(
     ]
     dataset.upload_predictions(model, predictions)
     return predictions
+
+
+def assert_partial_equality(expected, result, base_keys=[]):
```
jihan-yin, line 551: can we add a comment that this is checking partial equality in that all the keys from expected must be present and matching in the result? was a little confusing to me at first glance before i figured that out

jihan-yin (approved):

tests/helpers.py
jihan-yin: like the direction of partial equality

ardila: Added a comment.

PR #239, opened 2022-03-03
Changes for new video schema

cmpajot:
Removes `video_frame_url` argument for DatasetItem and instead requires video frames to be passed in as `image_url`. 

Accompanies video schema change [PR](https://github.com/scaleapi/scaleapi/pull/36689).

Will update read me docs to reflect this change once this is merged.


sasha-scale (commented):
breaking change is fine since to my knowledge not many customers using this right now. I think this PR is missing updates to the sphinx documentation though, and consider that you'll potentially have to update documentation in README as well

sasha-scale (approved):

PR #244, opened 2022-03-04
Fix rest of tests, use new autotag ids

gatli:
Here are a couple of fixes at once. There was a straggling ingest test that was failing similarly to what we had tested, essentially it is testing the returned messaging structure too stringently.

The CLI tests where hanging due to them trying to fetch data for 3500 datasets, after the great dataset purge we should be able to run those again.

I created a new autotag dataset with a test autotag and new slice there, this updates the old ids to the new ones.


ardila (approved):

ardila (approved):
Approving and merging since I think this does get us closer

PR #246, opened 2022-03-04
add predict_fn_or_cls as a bundle

ivan-khomyakov-sc:
Comes with https://github.com/scaleapi/models/pull/2684


nucleus/deploy/client.py
```diff
@@ -346,7 +364,12 @@ def sync_request(
         return resp["result_url"]
 
     def async_request(
-        self, endpoint_id: str, url: str, return_pickled: bool = True
+        self,
+        endpoint_id: str,
+        *,
+        url: Optional[str] = None,
```
seanshi-scale, line 370: Adding the kwargs functionality is also done in pr https://github.com/scaleapi/nucleus-python-client/pull/232, guess one of us will have a small merge conflict lol

nucleus/deploy/client.py
```diff
@@ -198,6 +210,8 @@ def create_model_endpoint(
         min_workers: int,
         max_workers: int,
         per_worker: int,
+        aws_role: str,
```
seanshi-scale, line 213: Don't think we should be exposing `aws_role` and `results_s3_bucket`, Yi talked about it at https://github.com/scaleapi/nucleus-python-client/pull/218#discussion_r809596648

There is functionality to add in aws_role and results_s3_bucket, by calling `client.register_endpoint_auth_decorator_fn` with a function 
```
def endpoint_auth_fn(payload):
    payload['aws_role'] = "<aws_role>"
    payload["results_s3_bucket"] = "<results_bucket>"
    return payload
```

yixu34: Yeah are you working off of an older branch?

nucleus/deploy/client.py
```diff
@@ -103,6 +107,7 @@ def create_model_bundle(
             model: Typically a trained Neural Network, e.g. a Pytorch module
             load_model_fn: Function that when run, loads a model, e.g. a Pytorch module
             load_predict_fn: Function that when called with model, returns a function that carries out inference
```
yixu34, line 109: It looks like the original type annotation of `load_predict_fn` was a `Callable`: `Callable[[DeployModel_T], Callable[[Any], Any]]`. In that case, maybe we don't need to create a new `predict_fn_or_cls` field, since `load_predict_fn` was already forwards compatible with the way you've specified the Bundle class with a `__call__` method?

ivan-khomyakov-sc: They have different usage:
a. predict_fn_or_cls
```
foo("woo") # returns inference result;
client.upload_bundle(predict_fn_or_cls=foo)
```

b. load_predict_fn
```
foo("woo") # returns inference result;

# incorrect:
client.upload_bundle(load_predict_fn=foo)

# correct:
def load_predict_fn(model):
    def predict(args):
          return foo(args)
    return predict
```

ivan-khomyakov-sc: I mean `predict_fn_or_cls` is a function/class that you can use directly in the service. `load_predict_fn ` on the other hand is a factory that produces a callable object.

nucleus/deploy/client.py
```diff
@@ -103,16 +103,20 @@ def register_endpoint_auth_decorator(self, endpoint_auth_decorator_fn):
...
+        predict_fn_or_cls: Optional[Callable[[Any], Any]] = None,
         requirements: Optional[List[str]] = None,
         model: Optional[DeployModel_T] = None,
         load_model_fn: Optional[Callable[[], DeployModel_T]] = None,
         bundle_url: Optional[str] = None,
     ) -> ModelBundle:
         """
         Grabs a s3 signed url and uploads a model bundle to Scale Deploy.
-        A model bundle consists of a "load_predict_fn" and exactly one of "model" or "load_model_fn", such that
+        A model bundle consists of a "predict_fn_or_cls" or "load_predict_fn" and exactly one of "model" or "load_model_fn", such that
```
seanshi-scale, line 119: I think the allowed combinations are  `{predict_fn_or_cls}, {load_predict_fn + model}, {load_predict_fn + load_model_fn}`, I'm not sure if it's completely clear from reading this. Maybe something like "A model bundle consists of {predict_fn_or_cls}, {load_predict_fn + model}, or {load_predict_fn + load_model_fn}.".

nucleus/deploy/client.py
```diff
@@ -139,11 +144,15 @@ def create_model_bundle(
...
+        check_args = [
+            predict_fn_or_cls is not None,
+            model is not None,
+            load_model_fn is not None,
+        ]
+
+        if sum(check_args) != 1:
             raise ValueError(
-                "Exactly one of model and load_model_fn should be non-None"
+                "Exactly one of `model` or `load_model_fn` or `predict_fn_or_cls` should be non-None"
```
seanshi-scale, line 155: Should we check that `load_predict_fn` doesn't get passed if a user passes `predict_fn_or_cls`?

nucleus/deploy/client.py
yixu34: Oh I see, the absence or presence of the `load` prefix can serve as a distinction as to whether or not it's directly used or not 👌🏻

nucleus/deploy/client.py
```diff
@@ -103,27 +103,32 @@ def register_endpoint_auth_decorator(self, endpoint_auth_decorator_fn):
     def create_model_bundle(
         self,
         model_bundle_name: str,
-        load_predict_fn: Callable[[DeployModel_T], Callable[[Any], Any]],
         env_params: Dict[str, str],
+        *,
+        load_predict_fn: Optional[
+            Callable[[DeployModel_T], Callable[[Any], Any]]
+        ] = None,
+        predict_fn_or_cls: Optional[Callable[[Any], Any]] = None,
```
yixu34, line 111: Overall I think this is fine for now, but we might as well start deprecating `load_predict_fn` later.

nucleus/deploy/client.py
```diff
@@ -161,22 +170,31 @@ def create_model_bundle(
...
+        elif model is not None:
+            bundle_func_1 = dict(model=model, load_predict_fn=load_predict_fn)
+            serialized_bundle = cloudpickle.dumps(bundle_func_1)
             bundle_metadata["load_predict_fn"] = inspect.getsource(
-                load_predict_fn
+                load_predict_fn  # type: ignore
             )
         else:
-            bundle = dict(
+            bundle_func_2 = dict(
```
yixu34, line 188: Can we just call both of these `bundle_func`? These branches are mutually exclusive, so even Python's funky dynamic scoping rules wouldn't be an issue.

nucleus/deploy/client.py
yixu34: Oh I guess same comment as https://github.com/scaleapi/models/pull/2684/files#r820019055, basically.

yixu34 (approved):

nucleus/deploy/client.py
ivan-khomyakov-sc: yeah, I did it for typing check that a linter performs. It complains that a variable `bundle` has different types in different branches, i.e dict/func/class_obj
I will play with the Optional[Dict...]

PR #248, opened 2022-03-08
Python client changes for new PUT endpoints/ functionality

seanshi-scale:
Testing:
Ran this against locally-hosted Gateway and it seems to send the celery request correctly


nucleus/deploy/client.py
```diff
@@ -426,6 +419,49 @@ def create_model_endpoint(
...
+        if gpus == 0 and gpu_type is not None:
+            logger.warning("GPU type setting %s will have no effect", gpu_type)
+            payload["gpu_type"] = None
+        for k, v in payload.copy().items():
+            if v is None:
+                del payload[k]
+        resp = self.connection.put(payload, f"{ENDPOINT_PATH}/{endpoint_name}")
+        endpoint_creation_task_id = resp.get("endpoint_creation_task_id", None)
+        logger.info(
+            "Endpoint creation task id is %s", endpoint_creation_task_id
```
yixu34, line 462: Endpoint edit task id?

nucleus/deploy/client.py
```diff
@@ -426,6 +419,49 @@ def create_model_endpoint(
...
             )
 
+    def edit_model_endpoint(
+        self,
+        endpoint_name: str,
+        model_bundle: Optional[ModelBundle] = None,
+        cpus: Optional[float] = None,
+        memory: Optional[str] = None,
+        gpus: Optional[int] = None,
+        min_workers: Optional[int] = None,
```
yixu34, line 429: Maybe this is clearer in the server side PR (I've so far mainly been looking at the abstractions and less so the semantics 😅 ) but if someone omits a field, do you read the existing value and ensure that it's preserved?

yixu34 (approved):

nucleus/deploy/client.py
seanshi-scale: Yup, that's the intended behavior

seanshi-scale:
We have our own client now, don't need this anymore

PR #252, opened 2022-03-09
A user can create a bundle that spans multiple files

ivan-khomyakov-sc:
https://app.shortcut.com/scaleai/story/396138/a-user-can-create-a-bundle-that-spans-multiple-files

Users can pickle bundles that rely on external modules. It's done by passing `globals()` value
```
client.create_model_bundle(..., globals_copy=globals())
```


pyproject.toml
```diff
@@ -49,7 +49,7 @@ click = ">=7.1.2,<9.0"  # NOTE: COLAB has 7.1.2 and has problems updating
 rich = "^10.15.2"
 shellingham = "^1.4.0"
 scikit-learn = ">=0.24.0"
-cloudpickle = "^1.6.0"
```
seanshi-scale, line 52: Does this mean we'll have to pin the version of cloudpickle inside of the async/sync inference endpoints to 2.0.0? Not entirely sure actually

poetry.lock
```diff
@@ -253,6 +253,46 @@ packaging = "*"
 six = ">=1.9.0"
 webencodings = "*"
```
seanshi-scale, line 255: ty for checking in poetry.lock lol (diego told me about this)

nucleus/deploy/client.py
```diff
@@ -109,6 +112,7 @@ def create_model_bundle(
         model: Optional[DeployModel_T] = None,
         load_model_fn: Optional[Callable[[], DeployModel_T]] = None,
         bundle_url: Optional[str] = None,
+        globals_copy: Optional[Dict[str, Any]] = None,
```
seanshi-scale, line 115: Is there a reason that the client has to pass in globals().copy(), or can we get the value of that from inside this function? I think if there isn't a reason for the client to pass this in then we shouldn't make them do that

nucleus/deploy/client.py
```diff
@@ -159,6 +165,15 @@ def create_model_bundle(
...
             )
 
+        # Prepare cloudpickle for external imports
+        if globals_copy:
+            for module in get_imports(globals_copy):
+                if module.__name__ == cloudpickle.__name__:
+                    # Avoid recursion
+                    # register_pickle_by_value does not work properly with itself
+                    continue
+                cloudpickle.register_pickle_by_value(module)
```
seanshi-scale, line 175: @yixu34 did this `cloudpickle.register_pickle_by_value` thing end up working? Thought it didnt?

pyproject.toml
ivan-khomyakov-sc: Nope!

Technically to my understanding and exploring the cloudpickle source code, you can use even simple pickle to deserialize

poetry.lock
ivan-khomyakov-sc: yeah, there was a weird bug with other packages. I guess we just need to call `poetry add <package>` instead of doing it manually in pyproject.toml

nucleus/deploy/client.py
ivan-khomyakov-sc: to my understanding no... There are some problems with using iterating `globals()` i.e. `for k, v in globals.items()` because size of globals changes while iterating. If you called globals from outside of the function and pass it as `globals_copy` then it's fine.

To answer you question: no it is not needed to `.copy()` it (I think). But I used this name to keep the same naming convention as in https://github.com/scaleapi/nucleus-python-client/blob/hmi-release/nucleus/deploy/find_packages.py#L323

nucleus/deploy/client.py
syandroo: hmm so it works just that the only limitation is you have to import the full module and outside the function?

ivan-khomyakov-sc: @syandroo given that they say on cloudpickle page: "Note that this feature is still experimental" I wouldn't be very hopeful :)

At least a simple use case of multiple .py files works fine.

I guess the next step is to try on one of the scale ml projects

PR #259, opened 2022-03-16
Improve scenario test interaction

pfmark:
- Scenario test getter was broken --> fixed
- Made the listing of scenario tests consistent with Nucleus dataset (property)


gatli (approved):
LGTM 👍

PR #261, opened 2022-03-16
Validate 3D metrics

Anirudh-Scale:
Added 3D IOU, BEV 2D IOU, and recall, precision metrics. Tested end-to-end with celery in `models` repo (see associated PR there).

### DIFF
--- a/nucleus/metrics/cuboid_metrics.py
+++ b/nucleus/metrics/cuboid_metrics.py
@@ -1,0 +1,197 @@
+import sys
+from abc import abstractmethod
+from typing import List
+
+from nucleus.annotation import AnnotationList, CuboidAnnotation
+from nucleus.prediction import PredictionList, CuboidPrediction
+
+from .base import Metric, ScalarResult
+from .filters import confidence_filter
+from .cuboid_utils import (
+    label_match_wrapper,
+    detection_iou,
+    recall_precision,
+)
+
+
+class CuboidMetric(Metric):
+    """Abstract class for metrics of cuboids.
+
+    The CuboidMetric class automatically filters incoming annotations and
+    predictions for only cuboid annotations. It also filters
+    predictions whose confidence is less than the provided confidence_threshold.
+    Finally, it provides support for enforcing matching labels. If
+    `enforce_label_match` is set to True, then annotations and predictions will
+    only be matched if they have the same label.
+
+    To create a new concrete CuboidMetric, override the `eval` function
+    with logic to define a metric between cuboid annotations and predictions.
+    """
+
+    def __init__(
+        self,
+        enforce_label_match: bool = False,
+        confidence_threshold: float = 0.0,
+    ):
+        """Initializes CuboidMetric abstract object.
+
+        Args:
+            enforce_label_match: whether to enforce that annotation and prediction labels must match. Default False
+            confidence_threshold: minimum confidence threshold for predictions. Must be in [0, 1]. Default 0.0
+        """
+        self.enforce_label_match = enforce_label_match
+        assert 0 <= confidence_threshold <= 1
+        self.confidence_threshold = confidence_threshold
+
+    @abstractmethod
+    def eval(
+        self,
+        annotations: List[CuboidAnnotation],
+        predictions: List[CuboidPrediction],
+    ) -> ScalarResult:
+        # Main evaluation function that subclasses must override.
+        pass
+
+    def aggregate_score(self, results: List[ScalarResult]) -> ScalarResult:  # type: ignore[override]
+        return ScalarResult.aggregate(results)
+
+    def __call__(
+        self, annotations: AnnotationList, predictions: PredictionList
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -45,7 +45,7 @@ click = ">=7.1.2,<9.0"  # NOTE: COLAB has 7.1.2 and has problems updating
 rich = "^10.15.2"
 shellingham = "^1.4.0"
 scikit-learn = ">=0.24.0"
-Shapely = ">=1.7.1"
+Shapely = "^1.8.1"
```
pfmark, line 48: @gatli didn't we run into problems with Shapely at some point?

nucleus/metrics/cuboid_metrics.py
```diff
@@ -0,0 +1,193 @@
...
+        self,
+        enforce_label_match: bool = True,
+        iou_threshold: float = 0.0,
+        confidence_threshold: float = 0.0,
+        birds_eye_view: bool = False,
+    ):
+        """Initializes CuboidIOU object.
+
+        Args:
+            enforce_label_match: whether to enforce that annotation and prediction labels must match. Defaults to False
```
pfmark, line 89: not consistent with the actual default value which is `True`. What do we want this to be?

nucleus/metrics/cuboid_metrics.py
```diff
@@ -0,0 +1,193 @@
...
+        self.iou_threshold = iou_threshold
+        self.birds_eye_view = birds_eye_view
+        super().__init__(enforce_label_match, confidence_threshold)
+
+    def eval(
+        self,
+        annotations: List[CuboidAnnotation],
+        predictions: List[CuboidPrediction],
+    ) -> ScalarResult:
+        iou_3d, iou_2d = detection_iou(
```
pfmark, line 106: just to avoid confusion, can we call `iou_2d` rather `iou_2d_bev`?

nucleus/metrics/cuboid_utils.py
```diff
@@ -0,0 +1,334 @@
...
+    mapping = []
+    for i, m in enumerate(iou_matrix.max(axis=1)):
+        if m >= threshold_in_overlap_ratio:
+            mapping.append((i, iou_matrix[i].argmax()))
+    return mapping
+
+
+def recall_precision(
+    prediction,
+    groundtruth,
```
pfmark, line 253: let's add typehints and also update the docstring for better usability

nucleus/metrics/cuboid_utils.py
```diff
@@ -0,0 +1,334 @@
...
+        "fp_sum": fp_sum,
+        "fn_sum": fn_sum,
+        "precision": tp_sum / (tp_sum + fp_sum),
+        "recall": tp_sum / (tp_sum + fn_sum),
+        "num_predicted": num_predicted,
+        "num_instances": num_instances,
+    }
+
+
+def detection_iou(prediction, groundtruth, threshold_in_overlap_ratio):
```
pfmark, line 310: let's also add typehints

pfmark (commented):
few smaller comments. In addition, do we have a way of easily testing the correctness of the metrics in a test unit test? I think that would be great!

pyproject.toml
gatli: Yup, we can't use it. It doesn't have wide enough binary support.

pfmark: So what should we do about this? @Anirudh-Scale how much do we rely in the shapely dependency? Any chance we can get around this? If not, @gatli how can we still make this work?

Anirudh-Scale: @pfmark I can implement a function to calculate the area intersection without Shapely, it's just going to be significantly slower. If that's a trade-off we're ok with, I can do that.

nucleus/metrics/cuboid_metrics.py
Anirudh-Scale: Good catch, will edit the comment.

nucleus/metrics/cuboid_metrics.py
```diff
@@ -0,0 +1,194 @@
...
+    `enforce_label_match` is set to True, then annotations and predictions will
+    only be matched if they have the same label.
+
+    To create a new concrete CuboidMetric, override the `eval` function
+    with logic to define a metric between cuboid annotations and predictions.
+    """
+
+    def __init__(
+        self,
+        enforce_label_match: bool = False,
```
harish-kamath, line 29: This argument (and maybe confidence_threshold?) should be required in the constructor, and if we want to set defaults then set them in the child classes.

pyproject.toml
phil-scale: There's already a function for polygon area intersection in the directory under `geometry.py`. I'd suggest using that since it looks like you're using shapely only for polygon intersection area.

sasha-scale (approved):

PR #263, opened 2022-03-18
Da item pagination

ardila:
* Add slice and dataset item paginated batch export
* Generalized paginated export utility
* Clean up existing code for item export (make it more general by parsing the full result with the class json parser, rather than custom logic)


nucleus/slice.py
```diff
@@ -103,10 +81,52 @@ def dataset_id(self):
...
+
+        """
+        try:
+            dataset_item_jsons = self._client.make_request(
+                {}, f"slice/{self.id}", requests_command=requests.get
+            )[
+                "dataset_items"
+            ]  # Unfortunately, we didn't use a standard value here, so not using a constant for the key
+            return [
+                DatasetItem.from_json(dataset_item_json)
```
drakejwong, line 123: this will need corresponding changes for tests that use `slice.items`

drakejwong (approved):
left some comments, but stamping to unblock!

also needs version bump

PR #265, opened 2022-03-24
adding some files

patrick-scale:


sasha-scale:
Random style comments:
In python, it is conventional to name variables and filenames in snake_case, so I suggest you rename your files to:
```
IngestDevelopmentScripts/TestScript.py -> ingest_development_dcripts/test_script.py
```
etc.

nucleus/transform.py
```diff
@@ -0,0 +1,275 @@
...
+    def __array__(self):
+        return self.matrix
+
+    def __getitem__(self, values):
+        return self.matrix.__getitem__(values)
+
+    def __add__(self, other):
+        return Transform(other) @ self
+
+    def __matmul__(self, other):
```
sasha-scale, line 263: fancy

nucleus/transform.py
```diff
@@ -0,0 +1,275 @@
+import numpy as np
+import transforms3d as t3d
+from pyquaternion import Quaternion
+
+
+class Transform:
```
sasha-scale, line 6: seems like the user can supply a lot of different values in the constructor. Do all class methods apply to all possible types of values?  if not, you may want to separate into different class types (maybe use inheritance?)

nucleus/sensorfusion.py
```diff
@@ -0,0 +1,138 @@
+import numpy as np
```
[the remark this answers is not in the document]
patrick-scale: Having trouble running this

nucleus/sensorfusion.py
```diff
@@ -0,0 +1,138 @@
...
+            **kwargs (Dict[str, :class:`RawPointCloud, :class: CameraCalibration`]): Mappings from sensor name
+              to pointcloud and camera calibrations. Each frame of a lidar scene must contain exactly one
+              pointcloud and any number of camera calibrations
+    '''
+    def __init__(self, dev_pose: Transform = None, **kwargs):
+        self.dev_pose = dev_pose
+        self.items = {}
+        for key, value in kwargs.items():
+            if isinstance(value, CamCalibration):
+                self.items[key] = copy.copy(value)
```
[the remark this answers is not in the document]
patrick-scale: Otherwise this basically passes by reference. We want to generate a xFrames camera calibrations and this will just continually update the same variable otherwise

nucleus/transform.py
patrick-scale: this would be the work of a 3d engineer

nucleus/sensorfusion.py
```diff
@@ -0,0 +1,138 @@
...
+        offset = self.frames[0].dev_pose.inverse
+        for frame in self.frames:
+            frame.dev_pose = offset @ frame.dev_pose
+
+    def apply_transforms(self, relative: bool = False):
+        if relative:
+            self.make_transforms_relative()
+        for frame in self.frames:
+            for item in frame.items:
+                if isinstance(frame.items[item], RawPointCloud):
```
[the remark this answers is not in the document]
patrick-scale: It could be a CameraCalibration object

nucleus/sensorfusion.py
```diff
@@ -0,0 +1,138 @@
...
+        :returns: List of points in world coordinates
+        :rtype: np.array
+        """
+        for item in self.items:
+            if isinstance(self.items[item], RawPointCloud):
+                return np.hstack(
+                    [
+                        self.dev_pose @ self.items[item][:, :3],
+                        self.items[item][:, 3:4],
+                        self.items[item][:, 4:5]
```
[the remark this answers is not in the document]
patrick-scale: It actually just will skip this step if the shape is < 5 (add nothing)

nucleus/sensorfusion.py
```diff
@@ -0,0 +1,138 @@
...
+    Args:
+            dev_pose (:class:`Transform`): World Coordinate transformation for frame
+            **kwargs (Dict[str, :class:`RawPointCloud, :class: CameraCalibration`]): Mappings from sensor name
+              to pointcloud and camera calibrations. Each frame of a lidar scene must contain exactly one
+              pointcloud and any number of camera calibrations
+    '''
+    def __init__(self, dev_pose: Transform = None, **kwargs):
+        self.dev_pose = dev_pose
+        self.items = {}
+        for key, value in kwargs.items():
```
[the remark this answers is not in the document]
patrick-scale: Agreed - but it's tough because there is a variable number of camera items that they may add to each frame dependent on their sensor suite

nucleus/sensorfusion.py
```diff
@@ -0,0 +1,138 @@
...
+    def add_frame(self, frame: RawFrame = None):
+        self.frames.append(frame)
+
+    def make_transforms_relative(self):
+        """Make all the frame transform relative to the first transform/frame. This will set the first transform to position (0,0,0) and heading (1,0,0,0)"""
+        offset = self.frames[0].dev_pose.inverse
+        for frame in self.frames:
+            frame.dev_pose = offset @ frame.dev_pose
+
+    def apply_transforms(self, relative: bool = False):
```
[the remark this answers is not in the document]
patrick-scale: Commited new changes to handle double transformation

ingest_development_scripts/testing_script.py
```diff
@@ -0,0 +1,43 @@
+from nucleus.sensorfusion import *
```
ardila, line 1: We do this elsewhere in the code but it's generally not good style, so good to avoid if possible

ardila, line 1: This is close to being a real unit test but

A) The test data you use isn't being checked in
B) There are no automatic checks for expected data


Since this is a first pass, fine to skip for now, just letting you know this is VERY close!

nucleus/sensorfusion_utils.py
```diff
@@ -0,0 +1,19 @@
+import open3d as o3d
```
ardila, line 1: This needs to be added as a project dependency using 

`poetry add`

I think it's as simple as

`poetry add open3d`

nucleus/transform.py
```diff
@@ -0,0 +1,275 @@
+import numpy as np
+import transforms3d as t3d
```
ardila, line 2: 2 more undeclared dependency that needs to be added so when users

pip install scale-nucleus

they won't get an error when running this file.

ardila (changes requested):
I think this is clean, self contained at a high level. Nice job patrick. I really like where you landed with the high level design of this. There are few concepts for users to learn and they appear to be easy to use.

We do need to add unit tests to ensure that it works as expected for our users. For example, right now users would get import errors, because some of the dependencies of this code have not been added to the project.

ingest_development_scripts/testing_script.py
patrick-scale: Will check this out!

PR #270, opened 2022-04-04
[Launch] add bundle_id for model creation

syandroo:
added the ability to link a bundle to the model project object in #34891, this adds it to the client so we can demo


nucleus/model.py
```diff
@@ -121,6 +123,7 @@ def from_json(cls, payload: dict, client):
             name=payload["name"],
             reference_id=payload["ref_id"],
             metadata=payload["metadata"] or None,
+            bundle_id=payload["bundle_id"] or None,
```
ivan-khomyakov-sc, line 126: I am curious, is `payload["smth"] or None` is justified, or `payload.get("smth", None)` would be better? Or wee need the code to break if bundle_id is not provided?

phil-scale:
The interface change looks reasonable to me, though I wonder if you want to use `bundle_id` or `bundle_name` - is the bundle name guaranteed to be unique per user?

nucleus/model.py
syandroo: hmm - interesting question, i think metadata is also not a required field but the thing that calls this defaults it to None also so this is really just a fallback

syandroo:
@phil-scale good q, it's actually mostly semantic since what we actually store is the bundle_name, (we used to refer to the bundle_name as bundle_id then moved some to bundle_name). on the user side we never expose the uuid that bundle or endpoint has, just their names.

syandroo:
i just wanted both the client and the API endpoint to have the same name, but i can change both to bundle_name if u think it's needed, just would require a lil bit more coordination on deploying.

nucleus/model.py
sasha-scale: +1 for payload.get("smth", None)

sasha-scale:
I agree with Phil, let's keep the user interface always using user defined names

nucleus/model.py
```diff
@@ -92,22 +92,24 @@ class Model:
     endpoint, using :meth:`NucleusClient.add_model`.
     """
 
-    def __init__(self, model_id, name, reference_id, metadata, client):
+    def __init__(self, model_id, name, reference_id, metadata, bundle_id, client):
```
seanshi-scale, line 95: Should we have a default `bundle_id=None` or is that something that Nucleus doesn't do? I guess if this isn't ever user-created it's probably fine to not have this default?

nucleus/model.py
```diff
@@ -187,3 +190,25 @@ def evaluate(self, scenario_test_names: List[str]) -> AsyncJob:
...
+
+            model.run("ds_123456")
+        
+        Args:
+            dataset_id: id of dataset to run inference on
+            job_id: nucleus job used to track async job progress
+            slice_id: (optional) id of slice of the dataset to run inference on
+        """
+        response = self._client.make_request(
+            {"dataset_id": dataset_id, "job_id": job_id, "slice_id": slice_id},
```
syandroo, line 209:
```suggestion
            {"dataset_id": dataset_id, "slice_id": slice_id},
```

nucleus/model.py
```diff
@@ -187,3 +190,25 @@ def evaluate(self, scenario_test_names: List[str]) -> AsyncJob:
             requests_command=requests.post,
         )
         return AsyncJob.from_json(response, self._client)
+    
+    def run(self, dataset_id: str, job_id: str, slice_id: Optional[str]) -> str:
```
syandroo, line 194:
```suggestion
    def run(self, dataset_id: str, slice_id: Optional[str]) -> str:
```

syandroo (commented):
need to commit from phone

yixu34:
Don't we want to base this off of the `hmi-release` branch? Or have we moved onto `master`?

seanshi-scale:
> Don't we want to base this off of the `hmi-release` branch? Or have we moved onto `master`?

This looks like client changes on the nucleus side to support Launch integration, so I think this goes on `master`, irrespective of whatever happens to `hmi-release`.

sasha-scale:
that's correct, OK to merge this into master (independent of HMI release branch)

seanshi-scale (approved):
I'll defer to someone on nucleus but this looks good to me

yixu34 (approved):

PR #274, opened 2022-04-13
Custom object embeddings with annotation uploads

jihan-yin:
Finally here! + 2 tests

### DIFF
--- a/nucleus/annotation.py
+++ b/nucleus/annotation.py
@@ -12,6 +12,7 @@ from .constants import (
     CATEGORY_TYPE,
     CUBOID_TYPE,
     DIMENSIONS_KEY,
+    EMBEDDING_VECTOR_KEY,
     GEOMETRY_KEY,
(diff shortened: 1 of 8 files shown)


nucleus/annotation.py
```diff
@@ -121,6 +123,9 @@ class BoxAnnotation(Annotation):  # pylint: disable=R0902
             attach to this annotation.  Strings, floats and ints are supported best
             by querying and insights features within Nucleus. For more details see
             our `metadata guide <https://nucleus.scale.com/docs/upload-metadata>`_.
+        embedding_vector (Optional[List]): Custom embedding vector for this object annotation.
```
ardila, line 126: Don't need types here, I think sphinx inserts them from the type annotation in the function signature.

ardila (approved):

ardila:
Will have to fix the backend to merge this?

PR #278, opened 2022-04-20
Add Keypoints type to annotations and predictions

flubstep:
Keypoints labels are collections of points similar to polygon or line labels. They are firstly defined as a list of points, but each point is intended to match up to a static nomenclature (e.g. "chin", "left_ear", "right_ear") and also include a description of the intended display of this label via a collection of line segments called a "skeleton". An example of a keypoints label are facial keypoints, which can be seen on this example of a synthetic face:

image


flubstep:
CC @drakejwong, still waiting on a server-side merge before the tests will run properly, but running this by early in order to make sure the code checks out.

nucleus/__init__.py
```diff
@@ -51,6 +51,7 @@
     BoxAnnotation,
     CategoryAnnotation,
     CuboidAnnotation,
+    KeypointsAnnotation,
```
drakejwong, line 54: we'll need to add this to `__all__` for it to start showing up in the docs ([preview here](https://scale-ai-nucleus-python-sdk--278.com.readthedocs.build/en/278/api/nucleus/index.html))

nucleus/__init__.py
```diff
@@ -123,6 +124,7 @@
     BoxPrediction,
     CategoryPrediction,
     CuboidPrediction,
+    KeypointsPrediction,
```
drakejwong, line 127: same here, we'll need to add this to `__all__` for it to start showing up in the docs ([preview here](https://scale-ai-nucleus-python-sdk--278.com.readthedocs.build/en/278/api/nucleus/index.html))

nucleus/annotation.py
```diff
@@ -364,6 +370,144 @@ def to_payload(self) -> dict:
...
+    keypoints: List[Keypoint]
+    names: List[str]
+    skeleton: List[List[int]]
+    reference_id: str
+    annotation_id: Optional[str] = None
+    metadata: Optional[Dict] = None
+    embedding_vector: Optional[list] = None
+
+    def __post_init__(self):
+        self.metadata = self.metadata if self.metadata else {}
```
drakejwong, line 467: nit
```suggestion
        self.metadata = self.metadata or {}
```

nucleus/annotation.py
```diff
@@ -364,6 +370,144 @@ def to_payload(self) -> dict:
...
+
+    ::
+
+        from nucleus import KeypointsAnnotation
+
+        keypoints = KeypointsAnnotation(
+            label="face",
+            keypoints=[Keypoint(100, 100), Keypoint(120, 120), Keypoint(visible=False), Keypoint(0, 0)],
+            names=["point1", "point2", "point3", "point4"],
+            skeleton=[[0, 1], [1, 2], [1, 3], [2, 4]],
```
drakejwong, line 429: not super familiar with typical OOP paradigms for keypoints, but filed a long term follow up to support string based skeletons like `[["point1", "point2"], ...]`

nucleus/annotation.py
```diff
@@ -364,6 +370,144 @@ def to_payload(self) -> dict:
...
+    those keypoints.
+
+    ::
+
+        from nucleus import KeypointsAnnotation
+
+        keypoints = KeypointsAnnotation(
+            label="face",
+            keypoints=[Keypoint(100, 100), Keypoint(120, 120), Keypoint(visible=False), Keypoint(0, 0)],
+            names=["point1", "point2", "point3", "point4"],
```
drakejwong, line 428: any reason not to embed this in the `Keypoint` object?

drakejwong (approved):
ty! left some small comments. as you mentioned we're also awaiting the backend merge but will stamp to unblock

nucleus/annotation.py
flubstep: Yeah, I thought about that a little bit but ended up sticking with this solution. The reason why is that the labels themselves are really only the keypoints (the output of the network in COCO format is x, y, v), and so the names and skeleton are static across the entire project. Which means two things:

1. It's similar to how the taxonomy and the labels are separate in category annotations
2. It's easiest just to define your keypoint names + skeleton once and then just add them as a keyword in the constructor, as opposed to having to zip() the keypoint names against the x,y coordinates

drakejwong: gotcha that makes sense, thanks!

PR #279, opened 2022-04-21
Use Shapely for 2D polygon calculations

gatli:
We decided to start using shapely again for 3d performance reasons. To be better prepared for semantic segmentation we want to start using shapely again for more stability on the 2D poly front.

@phil-scale do you see any problems with this revert?


phil-scale (approved):
LGTM - I ❤️ shapely!

PR #282, opened 2022-04-21
Add additional check to keypoints annotations

flubstep:
Assert that KeypointsAnnotation.names has all unique names, otherwise there may be issues with the rendering later


jean-lucas (approved):
LGTM

gatli:
Oops, that one slipped through the cracks on my end 😅

PR #288, opened 2022-04-26
Add id fields to Annotations and Predictions for traceability

gatli:
When loading annotations and predictions that have assigned IDs there was no way of searching for those in the frontend. This adds and ID field that is populated when loading Annotations and Predictions from server.


drakejwong:
Reiterating from Slack thread for visibility:
I'm not sure if we should pollute the constructor params, since it risks the user interpreting this as a value they can supply (e.g. when tab completing via the function signature).

What do you think of an `id` property — this also extends to `DatasetItem`, `Dataset`, `Slice`, `Model`, etc — rather than a constructor param? This way we can better handle the case where the object has not yet hit our backend to have an ID generated.

PR #290, opened 2022-04-27
Add shapely import guards to make sure tests run without extras

gatli:
This adds a bunch of import guards that make sure not installing shapely doesn't error out. 

The tests now run without shapely.


cmpajot:
Any ideas how I can uninstall shapely so that I can test this? Did some googling but couldn't find a straightforward way.

gatli:
You can do
```
poetry shell
pip uninstall Shapely
 pytest tests/metrics
```

cmpajot (approved):
LGTM

PR #292, opened 2022-05-02
Added all segmentation to poly metrics

gatli:
Adds metrics that converts segmentation masks into polygons and then uses our existing polygon metrics.

## TODO:

- [ ] Allow disabling of iou matching. We don't want to always match the highest iou object (a seg mask might be split into two polys).


jean-lucas (approved):
LGTM, code looks super clean, well-documented and structured 🌟 

Are the metric available for a user to call from the SDK, if so, should we add some notes in the CHANGELOG and up the version?

gatli:
> LGTM, code looks super clean, well-documented and structured 🌟
> 
> Are the metric available for a user to call from the SDK, if so, should we add some notes in the CHANGELOG and up the version?

Yup, changelog and bump!

PR #295, opened 2022-05-05
Check for duplicate Reference IDs on Frame and Scene creation

jean-lucas:
As title suggests.

The original call for duplicates was misplaced, and not being called.


nucleus/dataset.py
```diff
@@ -572,8 +575,6 @@ def append(
                 video_scenes, update, asynchronous
             )
 
-        check_for_duplicate_reference_ids(dataset_items)
-
```
jean-lucas, line 576: this was not being checked for Scenes + Video

nucleus/scene.py
```diff
@@ -38,16 +42,15 @@ class Frame:
     """
 
     def __init__(self, **kwargs):
-        self.items = {}
+        self.items: Dict[str, DatasetItem] = {}
         for key, value in kwargs.items():
-            self.items[key] = value
-
-    def __post_init__(self):
```
jean-lucas, line 45: __post_init__ does not work on default classes.

nucleus/dataset_item.py
```diff
@@ -195,13 +195,15 @@ def check_all_paths_remote(dataset_items: Sequence[DatasetItem]):
 def check_for_duplicate_reference_ids(dataset_items: Sequence[DatasetItem]):
     ref_ids = []
     for dataset_item in dataset_items:
-        if dataset_item.reference_id is not None:
-            ref_ids.append(dataset_item.reference_id)
+        if dataset_item.reference_id is None:
+            raise ValueError("Reference ID cannot be None")
```
drakejwong, line 199:
```suggestion
            raise ValueError(f"Reference ID cannot be None! Encountered DatasetItem with no reference ID:\n{dataset_item}")
```
maybe something like this to make it easier for the user to debug? assuming the stack trace won't show it

nucleus/scene.py
drakejwong: nice catch!!

drakejwong (approved):
ty! just one nit to improve error logging

PR #300, opened 2022-05-17
add method for exporting predictions

sasha-scale:


nucleus/__init__.py
```diff
@@ -815,7 +815,6 @@ def make_request(
         Returns:
             Response payload as JSON dict.
         """
-        print(payload, route)
```
jean-lucas, line 818: 🙏 finally haha i was looking for this

nucleus/slice.py
```diff
@@ -234,6 +234,38 @@ def items_and_annotations(
...
+                        "category": List[CategoryAnnotation],
+                    }
+                }]
+        """
+        api_payload = self._client.make_request(
+            payload=None,
+            route=f"slice/{self.id}/{model.id}/exportForTraining",
+            requests_command=requests.get,
+        )
+        print(api_payload)
```
jean-lucas, line 266:
```suggestion
```

nucleus/utils.py
```diff
@@ -187,7 +188,7 @@ def format_dataset_item_response(response: dict) -> dict:
     }
 
 
-def convert_export_payload(api_payload):
+def convert_export_payload(api_payload, model_id: str = None):
```
jean-lucas, line 191: Would make more sense to pass a boolean instead of nullable string here. Since we're not using the
value of `model_id` anywhere.

ie: `convert_export_payload(api_payload, has_predictions: bool = False):`

jean-lucas (commented):
looks good, could you also update the version + changelog?

nucleus/slice.py
```diff
@@ -234,6 +234,38 @@ def items_and_annotations(
         )
         return convert_export_payload(api_payload[EXPORTED_ROWS])
 
+    def items_and_predictions(
```
pfmark, line 237: In general I think it would be easier to use if we had the same interface for exporting items, predictions and annotations between a `slice` and a `dataset`. Right now the corresponding method in a dataset is `export_predictions(model)`. I'd make this method consistent with the dataset one. 

Just wondering in case we have a huge export, wouldn't we also run into timeout issues then with this request?

nucleus/slice.py
sasha-scale: 😆

jean-lucas (approved):
thanks for addressing the comments. LGTM once conflicts are resolved 👍

PR #301, opened 2022-05-18
[Launch] allow the creation of a bundle from the nucleus client

syandroo:
The main idea here is that we want to be able to create a bundle + a linked Nucleus model in one go, as we're trying to reduce the steps needed to created a "launch model" so to speak.

Most of the controversy is likely going to happen with the importing of the launch-client, but I think a discussion needs to happen on maintaining consistency between these two client's requirements anyway. This is because all users who are going to use the Nucleus <> Launch integration have to use both clients.

Test plan:
Screen Shot 2022-05-18 at 4 26 57 PM


syandroo:
need to test lol

nucleus/__init__.py
```diff
@@ -518,6 +519,73 @@ def create_model(
             client=self,
         )
 
+    def create_launch_model(
+        self,
+        name: str,
+        reference_id: str,
+        bundle_args: Dict[str, any],
```
syandroo, line 526: going to fix this

.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     docker:
       # Important: Don't change this otherwise we will stop testing the earliest
       # python version we have to support.
-      - image: python:3.6-buster
```
ardila, line 12: don't change this please

ardila: can add another CI job for 3.7 if you want though

pyproject.toml
```diff
@@ -32,27 +32,27 @@ documentation = "https://dashboard.scale.com/nucleus/docs/api"
...
 requests = "^2.23.0"
 tqdm = "^4.41.0"
-dataclasses = { version = "^0.7", python = "^3.6.1, <3.7" }
 aiohttp = "^3.7.4"
 nest-asyncio = "^1.5.1"
 pydantic = "^1.8.2"
 numpy = "^1.19.5"
 scipy = ">=1.4.1"  # NOTE: COLAB has 1.4.1 and has problems updating
 click = ">=7.1.2,<9.0"  # NOTE: COLAB has 7.1.2 and has problems updating
-rich = "^10.15.2"
```
ardila, line 45: try changing to >=

nucleus/__init__.py
```diff
@@ -48,6 +48,7 @@
 import tqdm
 import tqdm.notebook as tqdm_notebook
 
+from launch import LaunchClient
```
seanshi-scale, line 51: would it make sense to have all the launch integration stuff inside some `nucleus/launch` subdirectory, s.t. nothing in `nucleus` imports from `nucleus/launch`, but `nucleus/launch` stuff imports from nucleus?

and then separately, I know you can theoretically do some `pip install scale-nucleus[launch-integration]`, although we still need to keep packages in sync between the two libraries

pyproject.toml
```diff
@@ -32,27 +32,27 @@ documentation = "https://dashboard.scale.com/nucleus/docs/api"
...
+rich = "^12.0.0"
 shellingham = "^1.4.0"
 scikit-learn = ">=0.24.0"
 Shapely = { version = ">=1.8.0", optional = true }
 rasterio = "^1.2.10"
 Pillow = ">=8.3.1"
 s3fs = ">=2022.1.0"
+scale-launch = { git = "https://github.com/scaleapi/launch-python-client.git", python = ">3.7, <4.0.0" }
```
seanshi-scale, line 51: guess scale-launch isn't on pypi yet heh

nucleus/__init__.py
seanshi-scale, line 526: Should we directly expose things required for launch.create_model_bundle{_from_dir}()? (e.g. load_predict_fn/predict_fn/etc.)?
On second thought, that might make maintaining version compatibility really annoying

syandroo: yeah... i need to fix the any for linting but we'd have to make changes twice :/

nucleus/__init__.py
syandroo: if this is to prevent the requirements issues, i think i've gotten it to a relatively resolved state? with all the poetry wrangling lol

nucleus/__init__.py
```diff
@@ -518,6 +519,73 @@ def create_model(
...
+            for bundle in launch_client.list_model_bundles()
+        )
+
+        if bundle_exists or model_exists:
+            raise ModelCreationError(
+                "Bundle with the given name already exists, please try a different name"
+            )
+
+        kwargs = {
+            "model_bundle_name": name + "-nucleus-autogen",
```
syandroo, line 576: @seanshi-scale something i do wonder is how we should name this bundle... does a suffix make sense or should we just make it named the name passed in in general?

seanshi-scale: hmm this seems like an "api" question? I feel like I don't have too strong of an opinion but ideally something that says "created from nucleus" e.g. something in `bundle_metadata` (lol) or in the name feels like a good idea. (Also I feel like if clients have a deeper integration, adding a suffix might cause weird behavior)

nucleus/__init__.py
yixu34: Hm yeah this is interesting... in the process of trying to streamline one particular case, it does bring up the point of possibly writing 2N the number of client functions. Think it's probably fine for now? The nucleus client is just using the launch client's public functions anyway.

yixu34: Maybe another way of thinking about this: the only added value from doing Launch operations inside the Nucleus client is that you don't have to do a separate set of Nucleus operations afterwards. Just a broad thought to keep in mind as we're building this out.

nucleus/__init__.py
yixu34: Generally speaking, it doesn't matter how you identify the Nucleus-associated objects, as long as your client library functions can identify them again (i.e. you define the inverse operation). That said, all other things held equal, it'd probably be better to do this in a more structured way, which means that metadata (even though metadata itself is unstructured) is better than just a naming convention.

yixu34 (commented):
~Let's switch to metadata, and also get an approval from someone on the Nucleus side.~
Nevermind, don't use metadata. But do get Nucleus approval.

nucleus/__init__.py
yixu34: Another thought, assuming we put things into metadata: we may want to reserve a convention that metadata fields that have a certain prefix (e.g. `_` or `__` are "internal", aka they're used by us for bookkeeping for things like Nucleus identification).

syandroo: re: nucleus -> launch querying, we have a column in the nucleus database for `bundle_id`

launch -> nucleus querying _i think_ would not be needed, since any bundle created thru the nucleus client has an attached model

i like the __ ideal for internal too hmm

syandroo: also i'd prefer not stuffing things into internal metadata if possible, we are using relational dbs for a reason, if we need it can just make a new column called "generated_by" and have it contain the internal product that made the bundle / endpoint

yixu34: Ok chatted a bit about this offline, but: it sounds like we don't need to do any Launch -> Nucleus querying (i.e. we do not need to know if a given Launch bundle has a Nucleus model; instead, the existing Nucleus model's `bundle_id`/`bundle_name` column is sufficient). So that means that any sort of metadata isn't a factor here 👌🏻 

All that you're really trying to accomplish here is give the endpoint _some_ name; adding an autogenerated suffix is purely a convenient visual marker (if we needed more structured querying, we could use some sort of other column or metadata - but as per the above discussion, we don't even need that). I think what @seanshi-scale is pointing out is that there may be an edge case where the customer has some workflow that assumes a certain suffix, and there might be subtle conflicts if that suffix matches ours. I think the way to go about this is to just claim a certain suffix convention for ourselves *as part of the public API*, e.g. `-scale-autogen`. This should discourage people from adopting the same suffix, but even if they do, catastrophic results seem unlikely.

tl;dr - suffix sounds fine, just pick something you're comfortable committing to as part of the public API, with the understanding it's a "soft" requirement.

syandroo:
test failures aren't my fault it seems

seanshi-scale (approved):
lgtm as long as someone from nucleus approves

nucleus/__init__.py
yixu34: I think you're implicitly assuming that the user has an expectation that `<name>` is what's used for both the bundle and the endpoint. I don't think we need to worry about this, because:
1. We can publish this suffix convention as part of our API spec.
2. Orthogonally, also as part of the API docs, we should show users that they should `get` the resulting endpoint and its name from this function, instead of assuming a particular pattern.

PR #302, opened 2022-05-19
Add real semseg metrics

gatli:
This add segmentation mask metrics. I'll need a couple of more eyes on the implementation. The references come from here: https://github.com/wkentaro/pytorch-fcn/blob/main/torchfcn/utils.py

We definitely need some eyes on the different metrics. I'm actually not sure if I'm implementing just plain precision or average precision for example 😓 

The following parts warrant special attention:
- ` _metric_impl` part of each of the metrics, all of the calculation are based of the confusion matrix calculated with `_fast_hist`
- This adds new filter types since segmentation stores metadata in different objects. This re-uses most of the pre-existing filtering logic but now we allow filtering by segment metadata and values (i.e. labels, indexes). We don't apply any filtering on the image but rather filter the resulting confusion matrix to only take into account segments that weren't filtered out.

There is a lot of boilerplate code here with then configs etc. so I'd focus on the `_metric_impl`.


nucleus/metrics/segmentation_metrics.py
```diff
@@ -141,52 +192,47 @@ def __init__(
...
-        self, annotations: AnnotationList, predictions: PredictionList
-    ):
-        return self.metric(annotations, predictions)
+        with np.errstate(divide="ignore", invalid="ignore"):
+            iou = np.diag(confusion) / (
+                confusion.sum(axis=1)
+                + confusion.sum(axis=0)
+                - np.diag(confusion)
+            )
+        return ScalarResult(value=np.nanmean(iou), weight=annotation_img.size)  # type: ignore
```
sasha-scale, line 217: Correct me if I'm wrong, but this function reads to me as a pixel wise iou computation (i.e. there's only pixel-level matching, no instance level matching). As such, it seems unexpected to me that this is only matching by index and not by label. AKA imagine someone passes an instance segmentation mask into this function, and the segments in the GT mask are:

```
{
    "car": 1,
    "truck": 2,
    "grass": 3,
}
```
and the predicted mask has segments:
```
{
    "car": 1,
    "truck": 2,
    "grass": 3,
    "car": 4,
}
```

I think I would expect pixel locations with gt 1 and pred 4 to be considered a "match" .


Same argument applies for when the indices in mask vs. GT are ordered differently, i.e. pred mask:
```
{
    "car": 2,
    "truck": 1,
    "grass": 4,
    "car": 3,
}
```

nucleus/metrics/segmentation_metrics.py
```diff
@@ -306,41 +375,40 @@ def __init__(
...
+
+        confusion = self._calculate_confusion_matrix(
+            annotation, annotation_img, prediction, prediction_img
         )
 
-    def call_poly_metric(
-        self, annotations: AnnotationList, predictions: PredictionList
-    ):
-        return self.metric(annotations, predictions)
+        with np.errstate(divide="ignore", invalid="ignore"):
```
sasha-scale, line 395: similar comment to above, for pickle's functional requirements, I think they're looking for instance-level recall/precision rather than pixel-level recall precision.  I judge this from trying to read through the coco-evaluation code they pointed us to (which is EXTREMELY hard to follow): https://github.com/facebookresearch/detectron2/blob/523c4023bc79719d2fc3fe0181311f88ad35a271/detectron2/evaluation/coco_evaluation.py#L222

So a). curious if you agree with me that the coco_evaluator seems to be doing instance level matching, and b). if that is indeed the case, I think we need to inject an extra instance matching step here to meet their requirements

nucleus/metrics/filtering.py
```diff
@@ -182,11 +250,15 @@ class MetadataFilter(NamedTuple):
     ListOfAndJSONSerialized,
 ]
 
+DNFFieldOrMetadataFilters = List[
```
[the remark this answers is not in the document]
gatli: This is disjunctive normal form, the explanatory comment is with another structure. I'll add it here 🙂

nucleus/metrics/segmentation_metrics.py
```diff
@@ -6,23 +6,28 @@
...
-    PolygonMAP,
-    PolygonPrecision,
-    PolygonRecall,
-)
+from .metric_utils import compute_average_precision
+
+# pylint: disable=useless-super-delegation
+
+
+def _fast_hist(label_true, label_pred, n_class):
```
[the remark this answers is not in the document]
gatli: Happens with copy pasted code! 😅

gatli:
> re: question posed in PR description, you've implemented pixel-wise precision.
> 
> I'm actually not completely sure what the distinction is between precision and average-precision. According to [this article](https://stats.stackexchange.com/questions/353116/average-precision-vs-precision#:~:text=Average%20precision%20gives%20you%20average,considering%20any%20specific%20decision%20threshold.) average precision applies when you take the average across a variety of thresholds(which ignores instance information) so I think the distinction between precision and average-precision is not particularly relevant

Yeah, we should name this precision then since we don't have any threshold to average over.

nucleus/metrics/segmentation_metrics.py
sasha-scale: sounds good, I am happy to leave that as a TODO

sasha-scale:
> > re: question posed in PR description, you've implemented pixel-wise precision.
> > I'm actually not completely sure what the distinction is between precision and average-precision. According to [this article](https://stats.stackexchange.com/questions/353116/average-precision-vs-precision#:~:text=Average%20precision%20gives%20you%20average,considering%20any%20specific%20decision%20threshold.) average precision applies when you take the average across a variety of thresholds(which ignores instance information) so I think the distinction between precision and average-precision is not particularly relevant
> 
> Yeah, we should name this precision then since we don't have any threshold to average over.

agreed!

sasha-scale (approved):
LGTM!

PR #305, opened 2022-05-27
[Validate] Create custom evaluation functions

jean-lucas:


nucleus/validate/client.py
```diff
@@ -175,3 +179,30 @@ def evaluate_model_on_scenario_tests(
             f"validate/{model_id}/evaluate",
         )
         return AsyncJob.from_json(response, self.connection)
+
+    def create_custom_eval_function(
```
gatli, line 183: I think this naming is not clear enough for the intended functionality. We're allowing customer to define some sort of sink that they can funnel their own external metrics into.

We already have the notion of a custom written eval function by customers although it is not exposed. We need to make it much clearer that this is a placeholder evaluation function to externally provide data.

nucleus/validate/data_transfer_objects/eval_function.py
```diff
@@ -81,3 +82,24 @@ class GetEvalFunctions(ImmutableModel):
...
+
+    @validator("name")
+    def name_is_valid(cls, v):  # pylint: disable=no-self-argument
+        if " " in v:
+            raise ValueError(
+                f"No spaces allowed in an evaluation function name, got '{v}'"
+            )
+        if len(v) == 0 or len(v) > 255:
+            raise ValueError(
+                "Name of evaluation function must be between 1-255 characters long"
```
gatli, line 103: Nice ❤️

gatli (changes requested):
Like said in another comment: I'd not merge this until we clear up the difference between a custom function and a customer_calculated or `external_` function.

We should finish the full workflow of uploading the metrics as well ... that'll probably surface more problems with the current abstraction.

jean-lucas:
@gatli does it make more sense now? It stays away from the existing `CustomEvalFunction` logic that existed before.

gatli (approved):
Nice! This is a lot clearer 🙂

PR #308, opened 2022-06-02
Export Scale task info

drakejwong:
From Dataset or Slice, at either the item or scene level depending on the dataset.
See backend PR: https://github.com/scaleapi/scaleapi/pull/40840

Test Plan
---
Tested locally on prodclone with: scene dataset, scene slice, image dataset, image slice.
I don't think we can test this without linked tasks in Mongo :/ don't want to slow down our test suite by blocking on creating a linked task (e.g. send_to_labeling -> export_scale_task_info).
FWIW send_to_labeling tests are also disabled, and there are no tests for ingest_tasks :(


drakejwong:
merging; ci failure was on a flake

PR #310, opened 2022-06-02
[Launch] use pypi package for launch instead of git repo

syandroo:
see title


drakejwong (approved):

jean-lucas (approved):
thanks for fixing it so quickly

PR #312, opened 2022-06-08
Open up dependencies

gatli:
This opens up dependencies and limits what we install when we don't install `metrics`. This makes installation possible in COLAB. We were erroring out before.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ exclude = '''
 
 [tool.poetry]
 name = "scale-nucleus"
-version = "0.13.0"
+version = "0.13.1"
 description = "The official Python client library for Nucleus, the Data Platform for AI"
 license =  "MIT"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -48,8 +48,8 @@ scikit-learn = ">=0.24.0"
 Shapely = { version = ">=1.8.0", optional = true }
 rasterio = { version = "^1.2.10", optional = true }
 Pillow = ">=7.1.2"
-s3fs = ">=2022.1.0"
-scale-launch = { version = ">=0.1.0", python = ">=3.7,<4.0" , optional = true}
+s3fs = {version = ">=2021.9.0", optional = true }
+scale-launch = { version = ">=0.1.0", python = ">=3.7,<4.0" }
```
jean-lucas, line 52: does `scale-launch` depend on `s3fs`? If so, should it remain `optional=true`?

PR #315, opened 2022-06-09
Add test matrix for installation

gatli:
This adds a CircleCI test matrix that checks if installation is working properly for the base package.

I've tried my best for now but can't get Python 3.10 test to pass so I've removed it for now.


jean-lucas (approved):
This is great 👍

PR #320, opened 2022-06-20
Speed up test suite by using pytest-xdist for more parallelism and cache poetry depends

gatli:
Uses `pytest-xdist` to use parallelism per worker. Also uses the `python` orb to cache dependencies in the test phase. 

This seems to get us down to about ~6-8 min per run instead of around 15 min.


gatli:
FYI @phil-scale, @sasha-scale, @syandroo.

You can use these tricks to speed up some of your tests. Using the `python` orb specifically speeds up a lot of installation steps.

PR #327, opened 2022-07-06
add list_query_items method

drakejwong:
Add method to run structured queries via API.

See corresponding backend PR https://github.com/scaleapi/scaleapi/pull/41652.

Piggybacking on the static dataset with embeddings to run queries, since otherwise we'd have to wait for newly appended items to get indexed in ES before running getting reliable structured query results.


nucleus/dataset.py
```diff
@@ -1677,3 +1677,23 @@ def update_item_metadata(self, mapping: Dict[str, dict]):
             self.id, self._client, mapping, ExportMetadataType.DATASET_ITEMS
         )
         return mm.update()
+
+    def list_query_items(self, query: str) -> Iterable[DatasetItem]:
```
jean-lucas, line 1681: Think this is a bit cleaner: 
```suggestion
    def query_items(self, query: str) -> Iterable[DatasetItem]:
```

jean-lucas (approved):
looks good, just wondering why prefix the function name with `list_`

nucleus/dataset.py
drakejwong: I jotted down some thoughts on unified pagination [in this doc](https://docs.google.com/document/d/1jJRIkRG2RBV1Am10vB3hRbTUAJXcGV5czLx8J3CtiAU/edit#)!

Relevant point here being, imo we should have a more consistent naming scheme for these "paginated list export" type endpoints. e.g.
- `list_items`
- `list_item_annotations`
- `list_query_items` <- not sure though, agree it is a bit awkward

But yes given that there isn't much of a naming convention at the moment, `query_items` sftm 😅

PR #328, opened 2022-07-08
Default delete_annotation ref_ids to an empty list

anthonykrivonos:
We're deprecating the `None` value which deletes all annotations due to [this SEV](https://scaleapi.slack.com/archives/C01U1N13Q7N/p1657210698221589)


drakejwong (approved):
ty! tests failing because of https://github.com/scaleapi/nucleus-python-client/pull/327

PR #329, opened 2022-07-12
none value for float is ok

sasha-scale:
Attempting to parse None as float should return None.  For some reason this isn't caught by ValueError


jonathanfeng-scale (approved):
stamping in case you need it

jean-lucas (approved):
LGTM, can also change the `except` block to `except (ValueError, TypeError): ... `
But this is fine 👌 

Thanks for fixig

sasha-scale:
Thanks for the suggestion @jean-lucas - that's more elegant :)

PR #338, opened 2022-07-29
Running metadata_update with async flag

jean-lucas:
Relevant BE changes: https://github.com/scaleapi/scaleapi/pull/43803

Allows the user to run `update_item_metadata` and `update_scene_metadata` tasks with a background worker.

This was previously causing timeouts on large updates


nucleus/metadata_manager.py
```diff
@@ -26,11 +27,18 @@ def __init__(
...
         self.dataset_id = dataset_id
         self._client = client
         self.raw_mappings = raw_mappings
         self.level = level
+        self.asynchronous = asynchronous
+
+        if len(self.raw_mappings) > 500 and not self.asynchronous:
+            print(
+                "Consider running the metadata_update with `asynchronous=True`, to avoid timeouts."
+            )
```
jean-lucas, line 41: should we throw an error or just print the suggestion?

gatli: I'm not sure what the best way to handle this sync/async discrepancy. Ideally we'd have the default `async` since that should work all of the time, but maybe people are generally looking at small batches and want it to be more snappy.

Let's throw an error for now and add a try/except block around the make request that allows us to hint if we timeout for smaller sets.

gatli (approved):
LGTM 👍 

Let's change the print to an exception and add the try/catch with a hint as well.

PR #340, opened 2022-08-04
return raw response

jean-lucas:
Sometimes we want more fine control with the request object.

### DIFF
--- a/nucleus/__init__.py
+++ b/nucleus/__init__.py
@@ -1021,7 +1021,8 @@ class NucleusClient:
         payload: Optional[dict],
         route: str,
         requests_command=requests.post,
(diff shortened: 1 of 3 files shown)


nucleus/__init__.py
```diff
@@ -1021,7 +1021,8 @@ def make_request(
         payload: Optional[dict],
         route: str,
         requests_command=requests.post,
-    ) -> dict:
+        get_raw_response: bool = False,
```
gatli, line 1024: Nit
```suggestion
        return_raw_response: bool = False,
```

gatli (approved):
Nice! LGTM 👍

PR #342, opened 2022-08-09
fix scene export payload; bump circleci resource_class

drakejwong:
- Fix keys when parsing Scene.to_json payload back into Scene.from_json
- Bump CircleCI `resource_class` from `small` -> `medium`...we'll have to investigate whether there was a regression later


anthonykrivonos (approved):
thanks 🙌

PR #347, opened 2022-08-15
fix job.status() in tests and CircleCI caching

drakejwong:
- Fix `AsyncJob` status payload keys causing test failures
- Fix `AsyncJob` export test
- Fix `page_size` for `.items_and_annotatation_generator()`
- Change to simple dependency install step to fix CircleCI caching failures


shiftsayan (approved):

drakejwong:
FYI: `update_autotag` test will fail until https://github.com/scaleapi/scaleapi/pull/44834 lands, so I will force-merge assuming all other tests pass

PR #353, opened 2022-09-09
Change CircleCI tests to use cached lockfile

gatli:
Changed the CircleCI tests to again depend on a cached lockfile. The generated lockfile is stored in `poetry.circleci.lock` and is used for our CI testing. I've also added scripts for dependency resolving ... the lockfile takes forever to generate. We could probably add more hints to our `pyproject.toml`

Major changes:
- Use `poetry.lock` for installation. Installation is about 50 seconds right now.
- Re-instate using the CircleCI python orb for better caching. Fixed the problem with the cache key check that happened before.
- Dropped `poetry install` test matrix. In the past I've had major problem with different envs. I think we need to deal with dev env problems on a case by case basis.
- Improved client installation test matrix to include tests for `metrics` and `launch` extras

The locking is still taking forever. We should try to figure out what packages are the main culprits and try to hint better for the resolver.


jean-lucas (approved):
LGTM

PR #354, opened 2022-09-12
Fix model metrics test

shiftsayan:
The model metrics are not being used anywhere in the test.

This call is causing errored jobs on the backend because by the time the async job starts the test is completed and dataset yielded in the test fixture is deleted from the database.

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -91,8 +91,6 @@ def test_new_model_endpoints(CLIENT, dataset: Dataset):
 
     dataset.upload_predictions(model, predictions=predictions)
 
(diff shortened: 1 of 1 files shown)


gatli (changes requested):
This is being used as an integration test for the model_run_commit job so this  is letting us know when the pipeline is broken. I'd rather we wait on the job for it to finish.

gatli (approved):
Since apparently we can't currently track the state of the evaluation job we can merge this. @shiftsayan is going to add a task to add `AsyncJob` tracking to the calculate_evaluation_metrics such that we can properly test that it runs to end.

PR #362, opened 2022-09-23
[Launch] Add a utility for users to test their Launch+Nucleus bundles locally

seanshi-scale:
If `show_image=True`, pops up a new window with the image. Also just returns the image with the boxes/lines/polygons drawn on top.

If we want we can use matplotlib.pyplot.imshow to display the image, although it looks uglier inside of the jupyter notebook (and I'd assume uglier as well from a python script)

Jupyter nb: vvv

image

Previous version: vvv
image

### DIFF
--- a/nucleus/test_launch_integration.py
+++ b/nucleus/test_launch_integration.py
@@ -1,5 +1,5 @@
 import io
-from typing import Any, Callable
+from typing import Any, Callable, Dict, List, Set
 
 from PIL import Image, ImageDraw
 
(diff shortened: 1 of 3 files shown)


nucleus/test_launch_integration.py
```diff
@@ -0,0 +1,74 @@
+import io
+from typing import Any, Callable, Dict
+
+from PIL import Image, ImageDraw
+
+# From scaleapi/server/src/lib/select/api/types.ts
+_TOP_LEVEL_REQUIRED_KEYS = {"geometry", "type"}
+_TOP_LEVEL_OPTIONAL_KEYS = {"label", "confidence", "classPdf", "metadata"}
+# TODO idk which ones are right for nucleus
```
seanshi-scale, line 9: TODO: I'm not sure which keys make sense to include, I've semi-arbitrarily included these 4 for now

nucleus/test_launch_integration.py
```diff
@@ -0,0 +1,117 @@
...
+    x: float
+    y: float
+    width: float
+    height: float
+
+
+class BoxAnnotationModel(BaseModel, extra=Extra.forbid):
+    geometry: BoxGeometryModel
+    type: str
+    label: Optional[str] = None
```
seanshi-scale, line 21: TODO these keys aren't complete, I'm not entirely sure which ones it makes sense to include

syandroo (approved):
amazin

PR #363, opened 2022-09-23
Fix sync ingest for large number of requests

jean-lucas:
When trying to run local uploads with large number of items, ie: 1k+ the following error appeared:
`Cannot connect to host api.scale.com:443 ssl:default [nodename nor servname provided, or not known”`

This (to my understanding) was caused by flooding the network with too many simultaneous requests. 

It seems like the Semaphore that was previously implemented to control the `concurrency` was not controlling the POST requests properly. 

[[sc-594516]](https://app.shortcut.com/scaleai/story/594516)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to the [Nucleus Python Client](https://github.com/scaleapi/n
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.14.20](https://github.com/scaleapi/nucleus-python-client/releases/tag/v0.14.20) - 2022-09-23
+
+### Fixed
+- Local uploads are correctly batched and prevents flooding the network with requests
+
+
 ## [0.14.19](https://github.com/scaleapi/nucleus-python-client/releases/tag/v0.14.19) - 2022-08-26
 
 ### Added
--- a/nucleus/async_utils.py
+++ b/nucleus/async_utils.py
@@ -1,3 +1,9 @@
+import os
+import logging
+import warnings
+os.environ['PYTHONASYNCIODEBUG'] = '1'
+logging.basicConfig(level=logging.DEBUG)
+warnings.resetwarnings()
 import asyncio
 import time
 from dataclasses import dataclass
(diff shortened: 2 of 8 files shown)


nucleus/async_utils.py
```diff
@@ -27,16 +33,7 @@ class FileFormField:
...
-
-async def gather_with_concurrency(n, *tasks):
-    """Helper method to limit the concurrency when gathering the results from multiple tasks."""
-    semaphore = asyncio.Semaphore(n)
-
-    async def sem_task(task):
-        async with semaphore:
-            return await task
-
-    return await asyncio.gather(*(sem_task(task) for task in tasks))
```
jean-lucas, line 39: The POST requests were being fired before it acquired the Semaphore.

nucleus/async_utils.py
```diff
@@ -169,47 +154,47 @@ async def _post_form_data(
...
-                    response.status in RetryStrategy.statuses
-                    and sleep_time != -1
-                ):
-                    time.sleep(sleep_time)
-                    continue
-
-                if response.status == 503:
-                    raise TimeoutError(
-                        "The request to upload your max is timing out, please lower local_files_per_upload_request in your api call."
+    async with UPLOAD_SEMAPHORE:
```
jean-lucas, line 159: Only change here is the addition of the semaphore context

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to the [Nucleus Python Client](https://github.com/scaleapi/n
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.14.20](https://github.com/scaleapi/nucleus-python-client/releases/tag/v0.14.20) - 2022-09-23
+
+### Fixed
+- Local uploads are correctly batched and prevents flooding the network with requests
```
pfmark, line 11: Any estimate for up to which number / size of data things would work robustly now? Or is there basically no limit and it would just be annoyingly slow?

nucleus/dataset_item_uploader.py
```diff
@@ -127,15 +121,15 @@ def _process_append_requests_local(
             requests.append(request)
 
         progressbar = self._client.tqdm_bar(
-            total=len(requests), desc="Local file batches"
+            total=len(requests),
+            desc=f"Uploading {len(items)} items in {len(requests)} batches",
```
pfmark, line 125: This makes sense, but for me the progress bar does not update. It just remains at
Screen Shot 2022-09-23 at 16 48 51

pfmark (approved):
tested it with working internet now and works well for 10k items. Would be good to fix the progress bar, otherwise good to go from my side!

CHANGELOG.md
jean-lucas: > Any estimate for up to which number / size of data things would work robustly now? 

IMO it should robustly work for any number of item uploads, since the client is now correctly batching.

nucleus/dataset_item_uploader.py
jean-lucas: (discussed it offline)

PR #364, opened 2022-09-26
Pytest dataset fixture no teardown

drakejwong:
Leaving the teardown to pytest causes race conditions in CircleCI because of multiple workers. We will add a cron to instead clean up datasets this on a daily basis, outside of typical CI windows.


jihan-yin (approved):
much needed change, ty

PR #367, opened 2022-10-17
Speed up shell and CLI invocation with Nucleus installed

gatli:
Managing imports to late import stuff that is not necessary for most users.
- np, sklearn and shapely on the metrics side (which nobody but our internal services use)
- tqdm notebook is now only loaded in notebooks.
- Shell completion script is now generated once and sourced -> this speeds up shell init considerably and autocompletion

Timing:
```
pipx install tuna
poetry run python -X importtime cli/nu.py 2>tuna.log
tuna tuna.log
```
# Before
image

# After

image


jean-lucas (approved):
nice improvements 😎

gatli:
> nice improvements 😎

❤️

PR #372, opened 2022-11-04
[Validate] Scene-level scenario tests and custom metrics uploads

anthonykrivonos:


nucleus/validate/scenario_test.py
```diff
@@ -224,6 +243,7 @@ def upload_external_evaluation_results(
             "overall_metric": aggregate_weighted_sum / aggregate_weight,
             "model_id": model_id,
             "slice_id": self.slice_id,
+            "level": "item" if is_item_eval else "scene",
```
vayunalapati, line 246: could we use the EntityLevel enum for this, also it could be cleaner to remove is_item_eval and change the var to the level of the result (which would get rid of this if statement here)

vayunalapati (approved):

nucleus/validate/data_transfer_objects/scenario_test_evaluations.py
```diff
@@ -1,15 +1,27 @@
...
 
+    @root_validator()
+    def is_item_or_scene_provided(
+        cls, values
+    ):  # pylint: disable=no-self-argument
+        if (
+            values.get("item_ref_id") is None
+            and values.get("scene_ref_id") is None
+        ):
+            raise ValueError("Must provide either item_ref_id or scene_ref_id")
```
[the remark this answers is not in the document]
anthonykrivonos: Yeah that's a good idea

nucleus/validate/scenario_test.py
anthonykrivonos: Yep sg!

nucleus/validate/scenario_test.py
```diff
@@ -170,6 +171,12 @@ def get_items(self) -> List[DatasetItem]:
             DatasetItem.from_json(item) for item in response[DATASET_ITEMS_KEY]
         ]
 
+    def get_scenes(self) -> List[Scene]:
+        response = self.connection.get(
+            f"validate/scenario_test/{self.id}/scenes",
+        )
+        return [Scene.from_json(scene) for scene in response[SCENES_KEY]]
+
```
gatli, line 179: I don't like exposing this dichotomy to the end user. Since tests can really only be of one level we should keep the same interface for all items of a test.

gatli (changes requested):
We need to figure out how we want to expose that scenario tests can be of multiple levels.

I don't like the idea of adding 
```
get_items
get_scenes
get_objects
get_tracks
```
where each one of these returns an empty list for different enums of levels. Let's just keep the `get_items` interface and switch there what types of items are returned based on the level of the test.

nucleus/validate/scenario_test.py
```diff
@@ -162,10 +164,14 @@ def get_eval_history(self) -> List[ScenarioTestEvaluation]:
         ]
         return evaluations
 
-    def get_items(self) -> List[DatasetItem]:
+    def get_items(
+        self, level: EntityLevel = EntityLevel.ITEM
```
anthonykrivonos, line 168: Base this on the level of the slice

gatli (approved):
LGTM 👍

nucleus/validate/scenario_test.py
```diff
@@ -224,6 +244,7 @@ def upload_external_evaluation_results(
             "overall_metric": aggregate_weighted_sum / aggregate_weight,
             "model_id": model_id,
             "slice_id": self.slice_id,
+            "level": str(level),
```
anthonykrivonos, line 247: Do multiple inheritance for EntityLevel from `str` and `Enum` and just pass the enum

PR #373, opened 2022-11-07
Support for Tracks

anthonykrivonos:
Draft only, release version + tests to come


nucleus/dataset.py
```diff
@@ -1862,3 +1851,75 @@ def query_items(self, query: str) -> Iterable[DatasetItem]:
         )
         for item_json in json_generator:
             yield DatasetItem.from_json(item_json)
+
+    def create_track(
```
gatli, line 1855: I think this should rather be on the `Scene` object abstraction. This is a bit too far away from where this would be used in my opinion.

nucleus/dataset.py
```diff
@@ -1862,3 +1851,75 @@ def query_items(self, query: str) -> Iterable[DatasetItem]:
...
+            reference_id (str): The user-defined reference to the track.
+                Must be unique per dataset.
+            scene_reference_id (Optional[str]): The user-defined reference to the scene this track
+                pertains to.
+            metadata (Optional[dict]): An arbitrary dictionary of additional data about this track that can be stored
+                and retrieved.
+
+        Returns:
+            The created `Track` object.
+        """
```
gatli, line 1874: Are we expecting tracks on dataset items that aren't scenes? 

I think we should go for an immutable model for tracks. On creating a track should have references to all the objects belonging to it.

nucleus/dataset.py
```diff
@@ -1862,3 +1851,75 @@ def query_items(self, query: str) -> Iterable[DatasetItem]:
...
+                TRACK_REFERENCE_IDS_KEY: [
+                    track.reference_id for track in tracks
+                ]
+            },
+            route=f"dataset/{self.id}/track",
+            requests_command=requests.delete,
+        )
+
+    @property
+    def tracks(self) -> List[Track]:
```
gatli, line 1908: This on the other hand belongs on the dataset level 🙂

nucleus/track.py
```diff
@@ -0,0 +1,166 @@
...
+                else [],
+                PREDICTIONS_KEY: [pred.reference_id for pred in predictions]
+                if predictions
+                else [],
+            },
+            route=f"dataset/{self.dataset_id}/track/link",
+            requests_command=requests.post,
+        )
+
+    def unlink(
```
gatli, line 141: To reiterate my prior thought: I think we should try our best to make tracks immutable. A track probably has a level as well, it doesn't ever have annotation and predictions in the same track.

nucleus/track.py
```diff
@@ -0,0 +1,166 @@
...
+            METADATA_KEY: self.metadata,
+        }
+
+        return payload
+
+    def to_json(self) -> str:
+        """Serializes track object to schematized JSON string."""
+        return json.dumps(self.to_payload(), allow_nan=False)
+
+    def update(
```
gatli, line 67: Knowing our customers we probably need to support mutability for metadata 🙈 Or do you think we should rather go the immutability route here and make people update by deleting and replacing the track?

gatli (commented):
Creating a track and linking feels a little Janky in my head. I think the most common use case is actually going to be someone assigning a simple track index to an annotation/prediction when uploading rather than going the full create `Track` object.

On the other hand we need this pattern _as well_ to support uploading metadata to tracks.

But what do you think about changing the construction to always require you to pass in the reference ids of the items you are assigning to the track?

Any thoughts on implicitly creating simple tracks on upload of an Annotation/Prediction with an assigned track_index?

anthonykrivonos:
Seeking a preemptive review before I fix the tests up after the attached `scaleapi` PR is deployed

gatli (approved):
🙏

anthonykrivonos:
[This](https://github.com/scaleapi/scaleapi/pull/50654) should fix tests

PR #375, opened 2022-11-23
[Validate] Track-level metrics upload

anthonykrivonos:
Added support for uploading metrics at the track level from the SDK.

Tested everything in https://github.com/scaleapi/scaleapi/pull/50768. Will merge after the `scaleapi` changes are deployed and tests pass.


nucleus/__init__.py
```diff
@@ -178,7 +178,7 @@ def __init__(
             import tqdm.notebook as tqdm_notebook
 
             self.tqdm_bar = tqdm_notebook.tqdm
-        self._connection = Connection(self.api_key, self.endpoint)
+        self.connection = Connection(self.api_key, self.endpoint)
```
anthonykrivonos, line 181: Tricky design decision here. `Track` needs access to at least a `Connection` object to be able to call the API to upload track metadata, preferably `NucleusClient` access for consistency with other classes like `Model`. However, a `Validate` instance is created from `NucleusClient`, and so attempting to import `Track` from `ScenarioTest`, from where it is created in `get_items`, will cause a circular import. The lessermost evil was to update `Track.from_json` to take a `Connection` instead of a `NucleusClient`.

The tradeoff here was exposing `connection` as a public property of `NucleusClient` so to access this property when creating `Track`s in cases when only a `NucleusClient` is available. This should be safe, since a `Connection` consists of two public properties––`api_key` and `endpoint`.

Lmk if there are any issues with this approach.

nucleus/validate/scenario_test.py
```diff
@@ -178,6 +185,11 @@ def get_items(
         response = self.connection.get(
             f"validate/scenario_test/{self.id}/items",
         )
+        if level == EntityLevel.TRACK:
+            return [
+                Track.from_json(track, connection=self.connection)
```
anthonykrivonos, line 190: This line is the reason for the long justification above 😤

tests/validate/test_scenario_test.py
```diff
@@ -54,43 +55,92 @@ def test_scenario_test_get_dataset_items(
     test_slice,
     slice_items,
 ):
+    # Arrange
```
anthonykrivonos, line 58: Refactored the `test_scenario_test_get_*` methods via [AAA-C](https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80) bc of my OCD

nucleus/__init__.py
gatli: Yeah this is fine.

I think passing around a `connection` is a newer approach than passing the client. I actually don't see any reason for a class having to have full access to the client. We should probably refactor everything to pass around the connection object itself.

I don't see any harm in doing things this way so ... let's go ahead 🙂

nucleus/validate/constants.py
```diff
@@ -25,5 +25,6 @@ class ThresholdComparison(str, Enum):
 class EntityLevel(str, Enum):
     """Level for evaluation functions and unit tests."""
```
gatli, line 26: Let's elaborate in this docstring what the level really means.

nucleus/validate/scenario_test.py
gatli: Haha, the struggle is real 🥇

tests/validate/test_scenario_test.py
gatli: Feel free to drop the link next to one of the comments 👍

gatli (approved):
LGTM 👍

tests/validate/test_scenario_test.py
anthonykrivonos: I think it's ok, pretty common paradigm and future tests can use the structure of these functions as inspiration. Don't feel too fond about linking to Medium lmao

PR #379, opened 2023-01-16
[Nucleus] Get job status

jean-lucas:
### Changed
- Better filter tuning of `client.list_jobs(args)` method

### Added
- Dataset method to filter jobs, and statistics on running jobs
Example:
```python
>>> client = nucleus.NucleusClient(API_KEY)
>>> ds = client.get_dataset(ds_id)
>>> ds.jobs(show_completed=True, stats_only=True)
{'autotagInference': {'Cancelled': 1, 'Completed': 11},
 'modelRunCommit': {'Completed': 7, 'Errored_Server': 1, 'Running': 1},
 'sliceQuery': {'Completed': 40, 'Running': 2}}
```

Detailed Example
```python
>>> from nucleus.job import CustomerJobTypes
>>> client = nucleus.NucleusClient(API_KEY)
>>> ds = client.get_dataset(ds_id)
>>> from_date = "2022-12-20"; to_date = "2023-01-15"
>>> job_types = [CustomerJobTypes.MODEL_INFERENCE_RUN, CustomerJobTypes.UPLOAD_DATASET_ITEMS]
>>> ds.jobs(
  from_date=from_date,
  to_date=to_date,
  show_completed=True,
  job_types=job_types,
  limit=150
)
# ... returns list of AsyncJob objects
```


nucleus/__init__.py
```diff
@@ -251,23 +254,40 @@ def list_datasets(self) -> Dict[str, Union[str, List[str]]]:
         return self.make_request({}, "dataset/", requests.get)
 
     def list_jobs(
-        self, show_completed=None, date_limit=None
```
gatli, line 254: Add a warning that date_limit isn't working any more for good measure?

nucleus/__init__.py
```diff
@@ -251,23 +254,40 @@ def list_datasets(self) -> Dict[str, Union[str, List[str]]]:
...
-              running jobs. Default behavior is False.
-            date_limit: Only fetch jobs that were started after this date. Default
-              behavior is 2 weeks prior to the current date.
-
-        Returns:
-            List[:class:`AsyncJob`]: List of running asynchronous jobs
-            associated with the client API key.
+            job_types: Filter on set of job types, if None, fetch all types
+            from_date: beginning of date range filter
+            to_date: end of date range filter
```
gatli, line 270: Can you show an example of a valid string?

gatli (approved):
LGTM 👍

nucleus/__init__.py
jean-lucas: How would we do that?

If someone now tries to call `list_jobs(date_limit=foo)` then it will give `TypeError`.
Is there a clean way to handle this?

Only way I can think of, is to introduce a temporary 2nd function.

pfmark (commented):
would this ever show errored jobs? Because those are neither completed nor running

jean-lucas:
> would this ever show errored jobs? Because those are neither completed nor running

yes, it will fetch all job statues. Or you can set `show_completed=False` to skip those

pfmark:
> Or you can set `show_completed=False` to skip those

I think this is not the most intuitive interface. Would rename at least to `filter_only_completed`

PR #381, opened 2023-02-15
Limit async append jobs to 1000 items

jean-lucas:
The server does not batch incoming requests.
As such, we limit each job request to have at most 1000 items


gatli (commented):
Hmm, I think this error should originate from the server. Most people aren't adamant about updating the client such that it won't affect them. Hopefully we'll fix this in the future so ideally we should be throwing from the server side.

PR #391, opened 2023-09-14
add raw export for object slice

pfmark:
Can export json summaries of object slices containing predictions, annotations, item and scene level data per object in a slice.


jean-lucas (approved):
once docstring is updated, LGTM

jean-lucas (approved):
lgtm

PR #394, opened 2023-09-18
Add support for asynchronous embeddings export

ntamas92:


nucleus/async_job.py
```diff
@@ -131,6 +131,27 @@ def from_json(cls, payload: dict, client):
...
+                [{
+                    "reference_id": str,
+                    "embedding_vector": List[float]
+                }]
+        """
+        status = self.status()
+
+        if status["status"] != "Completed":
+            raise JobError(status, self)
+
```
jean-lucas, line 151: why raise a `JobError` if the job is not completed? perhaps its still running?

ntamas92: My thought process was that the usage pattern would be the following:

```python
exportJob = dataset.export_embeddings()
exportJob.sleep_until_complete(False)
result = exportJob.result_urls()
```

We could just wait for the result urls inside `result_urls()` also, but then I'd highlight it somehow that obtaining the results could run for a long time.

.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           command: |
             poetry run black --check .
       - run:
-          name: Ruff Lint Check # See pyproject.tooml [tool.ruff]
+          name: Ruff Lint Check # See pyproject.toml [tool.ruff]
```
gatli, line 42: Merci 🙏

nucleus/async_job.py
jean-lucas: Alright, that makes sense, didn't noticed the AsyncJob inheritence.
This is a neat idea, to have customized job result classes

jean-lucas (approved):
LGTM 👍

nucleus/async_job.py
gatli: Maybe we should add a `wait_for_completion` parameter. It might even be the default to wait for the job to complete.

nucleus/async_job.py
```diff
@@ -131,6 +131,27 @@ def from_json(cls, payload: dict, client):
         )
 
 
+class EmbeddingsExportJob(AsyncJob):
+    def result_urls(self) -> List[str]:
```
gatli, line 135: I'm wondering how you would instantiate this in another process. I think we need a classmethod `from_id` that would allow you to spin this up in one environment and then poll in another just from the `job_id`

gatli (commented):
Look good to me! Let's address the instantiation of the EmbeddingsExportJob (and for that matter any AsyncJob) such that we can let people trigger this in one process and poll in another.

gatli:
image
We really need to address this flakiness 😓

nucleus/async_job.py
ntamas92: Good point, that would be used through the `NucleusClient.list_jobs` method though right? So something like this:

```
jobs = NucleusClient.list_jobs()
export_job = EmbeddingsExportJob.from_job_id(jobs[0].job_id)
```

nucleus/async_job.py
ntamas92: Good idea, let's do that

PR #396, opened 2023-09-28
Image Chip Generator

vayunalapati:
Create an image and boxannotation chip generator.

As part of ML training workflows its preferable to train on chips rather than entire images for model performance and training speed. For this we grab the image url from Nucleus, generate a chip, and save it either locally or to S3.


nucleus/utils.py
```diff
@@ -422,3 +434,87 @@ def paginate_generator(
             yield json_value
         if not next_token:
             break
+
+def fetch_image(s3_url: str) -> Image.Image:
+    s3_bucket = s3_url.split("//")[-1].split("/")[0]
```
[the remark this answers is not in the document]
robert-perrotta: Or you could follow the lead of [aws-cli](https://github.com/aws/aws-cli/blob/b2ffab1bf2bdff4c7212e3680d47fed362053a2a/awscli/customizations/s3/utils.py#L238)

nucleus/utils.py
```diff
@@ -422,3 +434,87 @@ def paginate_generator(
...
+        boto3.resource("s3").Bucket(s3_bucket).Object(s3_output).get()["Body"]
+    )
+    return image
+
+def fetch_chip(s3_prefix: str) -> tuple[str, str]:
+    s3_bucket = s3_prefix.split("//")[-1].split("/")[0]
+    s3_output = s3_prefix.split("//")[-1].split(s3_bucket)[-1][1:]
+    try:
+        boto3.resource("s3").Bucket(s3_bucket).Object(s3_output + '.jpeg').load()
+    except:
```
robert-perrotta, line 451: You should filter on exception type so this doesn't hide other bugs/failures. At the very least, `except Exception` will still raise `KeyboardInterrupt` so you can still ctrl+c to quit.

nucleus/utils.py
```diff
@@ -422,3 +434,87 @@ def paginate_generator(
...
+    try:
+        boto3.resource("s3").Bucket(s3_bucket).Object(s3_output + '.jpeg').upload_fileobj(byteio)
+    except Exception as err:
+        print("hi", err)
+        return None, None
+    if len(chipped_annotations) == 0:
+        return s3_prefix + '.jpeg', None
+    try:
+        boto3.resource("s3").Bucket(s3_bucket).Object(s3_output + '.json').put(Body=json.dumps(chipped_annotations))
+    except Exception as err:
```
robert-perrotta, line 474: If this raises an error, you can end up finding the image chip without annotations, which will be treated as though there are no annotations for that chip. You could either write empty annotation files or just load the annotations first to work around this.

nucleus/utils.py
```diff
@@ -422,3 +434,87 @@ def paginate_generator(
...
+        print("sup", err)
+        return None, None
+    return s3_prefix + '.jpeg', s3_prefix + '.json'
+
+def generate_offsets(w: int, h: int, chip_size: int, stride_size: int):
+    xs = np.arange(0, w - stride_size, chip_size - stride_size)
```
robert-perrotta, line 480: I think you want `np.arange(0, w - stride_size, stride_size)`

nucleus/utils.py
```diff
@@ -422,3 +434,87 @@ def paginate_generator(
...
+        annotations = annotations.loc[annotations["area"] > 0]
+        annotations = (
+            annotations.groupby(["label", "x", "y", "width", "height"]).last().reset_index()
+        )
+    else:
+        annotations = pd.DataFrame([])
+    annotation_units = []
+    for _, annotation in annotations.iterrows():
+        annotation_units.append({
+            LABEL_KEY: annotation[LABEL_KEY],
```
robert-perrotta, line 510: The keys for `annotation` are used and/or set on the lines above and probably should be literals here. For example `LABEL_KEY: annotation["label"]`

nucleus/dataset.py
```diff
@@ -1414,10 +1429,63 @@ def items_and_annotation_generator(
...
         )
         for data in json_generator:
             for ia in convert_export_payload([data], has_predictions=False):
                 yield ia
+    
+    def items_and_annotation_chip_generator(
+        self,
+        chip_size: int,
+        stride_size: int,
+        s3_directory: str,
```
robert-perrotta, line 1442: Caching these results to S3 makes a lot of sense, but should it be optional?

nucleus/utils.py
```diff
@@ -422,3 +434,87 @@ def paginate_generator(
...
+        boto3.resource("s3").Bucket(s3_bucket).Object(s3_output + '.json').put(Body=json.dumps(chipped_annotations))
+    except Exception as err:
+        print("sup", err)
+        return None, None
+    return s3_prefix + '.jpeg', s3_prefix + '.json'
+
+def generate_offsets(w: int, h: int, chip_size: int, stride_size: int):
+    xs = np.arange(0, w - stride_size, chip_size - stride_size)
+    ys = np.arange(0, h - stride_size, chip_size - stride_size)
+    if len(xs) > 1:
```
robert-perrotta, line 482: I like this feature (shrink the stride_size so the last chip ends at the end of the image instead of extending beyond it) but will everyone? It should probably be mentioned in the docstring at least. You could also make it optional but then you have to decide if you want to drop the last (partial) chip or pad it with something.

nucleus/utils.py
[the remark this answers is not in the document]
robert-perrotta: Oh wow, we use that pattern all over the place. I guess it hasn't caused too many issues. All the same, there are more robust options.

robert-perrotta (commented):
Looking good, Vayun! I left a handful of comments, most minor. The biggest one is that I think the S3 caching should probably be optional.

nucleus/dataset.py
```diff
@@ -1414,10 +1429,63 @@ def items_and_annotation_generator(
...
+            url = data[ITEM_KEY][PROCESSED_URL_KEY]
+            image = fetch_image(url)
+            w, h = image.size
+            url = re.sub('[/.:]', '', url)
+            annotations = data[BOX_TYPE]
+            for x0, y0 in generate_offsets(w, h, chip_size, stride_size):
+                reference_id = f"{s3_directory}{chip_size}/{stride_size}/{url}_{x0}_{y0}"
+                chipped_image_url, chipped_annotation_url = fetch_chip(reference_id)
+                if chipped_image_url: 
+                    yield {IMAGE_URL_KEY: chipped_image_url, ANNOTATION_LOCATION_KEY: chipped_annotation_url}
```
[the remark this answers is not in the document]
vayunalapati: oh does yield not return the function early?

nucleus/utils.py
robert-perrotta: The variable name `stride_size` is a little confusing here but I saw in your demo this was the intended behavior.

nucleus/dataset.py
```diff
@@ -1414,11 +1429,76 @@ def items_and_annotation_generator(
...
         )
         for data in json_generator:
             for ia in convert_export_payload([data], has_predictions=False):
                 yield ia
 
+    def items_and_annotation_chip_generator(
+        self,
+        chip_size: int,
+        stride_size: int,
+        cache_directory: str,
```
robert-perrotta, line 1442: I think this should be optional

nucleus/dataset.py
```diff
@@ -1414,11 +1429,76 @@ def items_and_annotation_generator(
...
+                x1 = x0 + chip_size
+                y1 = y0 + chip_size
+                chipped_image = image.crop((x0, y0, x1, y1))
+                chipped_annotations = chip_annotations(
+                    annotations, x0, y0, x1, y1
+                )
+                chipped_image_loc, chipped_annotation_loc = write_chip(
+                    ref_id, chipped_image, chipped_annotations
+                )
+                yield {
```
robert-perrotta, line 1497: I would expect this function to generate the image chips and annotations themselves, rather than the locations. Or maybe both.

nucleus/utils.py
```diff
@@ -422,3 +437,118 @@ def paginate_generator(
...
+    if len(ys) > 1:
+        ys = np.round(ys * (h - chip_size) / ys[-1]).astype(int)
+    yield from product(ys, xs)
+
+
+def chip_annotations(data, x0: int, y0: int, x1: int, y1: int):
+    annotations = []
+    for annotation in data:
+        geometry = annotation[GEOMETRY_KEY].copy()
+        geometry[W_KEY] = geometry[X_KEY] + geometry[WIDTH_KEY]
```
robert-perrotta, line 532: I think the variable names like `W_KEY` are confusing. Maybe something based on x1, x2 or left, right etc?

robert-perrotta (commented):
This is looking really good! I nit-picked a few things and also had a question about the inputs and outputs of `items_and_annotation_chip_generator`. Where possible, I think it would be best to model it after `items_and_annotation_generator` that way it could be a drop in replacement for the latter.

nucleus/dataset.py
vayunalapati: If we are writing the images to a location, which I detail further below, it should be necessary

nucleus/dataset.py
vayunalapati: in terms of expectations we do provide in the doc string the output format

the reason I went forward with a location based approach is since this is going to be used on training datasets, users would have to implement this logic themselves (since there's not a great way of storing all of these annotations and images in memory) instead of allowing this to do it all for them.
we would also need to return the pillow image in a dictionary which does not seem very safe

nucleus/utils.py
vayunalapati: updated

nucleus/dataset.py
robert-perrotta: But I mean the whole caching to file feature should be optional. I think this feature would be useful in some instances in which I wouldn't want it storing the chips anywhere, just generating them transiently.

nucleus/dataset.py
robert-perrotta: Since you are yielding them, they never need to be stored in memory as a whole, and users still get all the benefits of caching if they just call this method with the same cache_directory etc. so they wouldn't lose any of that information.

We have

```py
def items_and_annotation_generator(
        self,
        query: Optional[str] = None,
    ) -> Iterable[Dict[str, Union[DatasetItem, Dict[str, List[Annotation]]]]]:
    ...
```
and it would be best if we also had (pseudocode)
```py
def items_and_annotation_chip_generator(
        self,
        chip_size: int,
        stride_size: int,
        cache_directory: Optional[str] = None,
        query: Optional[str] = None,
    ) -> Iterable[Dict[str, Union[DatasetItem, Dict[str, List[Annotation]]]]]:
    for item in items_and_annotation_generator(query):
        ...
        for x0, y0 in generate_offsets(w, h, chip_size, stride_size):
            try:
                chipped_image, chipped_annotations = fetch_chip(ref_id)
            except NotInCache:
                chipped_image, chipped_annotations = make_chip(ref_id, ...)
                write_chip(ref_id, chipped_image, chipped_annotations)
            yield {...}  # dict formatted like items_and_annotation_generator
```
you already have all the pieces to make that happen, and I think this interface would be easier to use.

nucleus/dataset.py
```diff
@@ -1414,11 +1429,76 @@ def items_and_annotation_generator(
...
+            chip=True,
+        )
+        for item in json_generator:
+            image = fetch_image(item[ITEM_KEY][PROCESSED_URL_KEY])
+            w, h = image.size
+            annotations = item[BOX_TYPE]
+            item_ref_id = item[ITEM_KEY][REFERENCE_ID_KEY]
+            for x0, y0 in generate_offsets(w, h, chip_size, stride_size):
+                x0, y0 = int(x0), int(y0)
+                ref_id = f"{cache_directory}/{chip_size}/{stride_size}/{item_ref_id}_{x0}_{y0}"
```
bdnorman, line 1480: I think this can probably just be `f"{cache_directory}/{item_ref_id}_{x0}_{y0}`. In a world in which we we have one dataset of chip_size=1024 and stride_size=512 and then another dataset of chip_size=512 and stride_size=512, there will be duplicate chips so we'd want to be able utilize the available cache from one to the other if possible

nucleus/dataset.py
```diff
@@ -1414,11 +1429,76 @@ def items_and_annotation_generator(
...
+        Returns:
+            Generator where each element is a dict containing the location of the image chip (jpeg) and its annotations (json).
+            ::
+
+                Iterable[{
+                    "image_location": str,
+                    "annotation_location": str
+                }]
+        """
+        json_generator = paginate_generator(
```
bdnorman, line 1465: Might be worth doing an info log if stride_size > chip_size then you're going to be missing out on some data

nucleus/dataset.py
bdnorman: agree with Robert here

nucleus/dataset.py
vayunalapati: in that case I think it should be `f"{cache_directory}/{item_ref_id}_{x0}_{y0}_{x1}_{y1}` otherwise the top left of the image would be the same in both cases which is not correct

nucleus/dataset.py
vayunalapati: sounds good I'll go ahead and do that

nucleus/dataset.py
vayunalapati: we are not able to return items in the way you would like to receive these items @robert-perrotta @bdnorman since once we modify the item, we are no longer returning `DatasetItem`s but instead raw images and its corresponding annotations. with this in mind I believe it's best to stick with the current approach

bdnorman: Synced with Vayun, I think this is fine for now and can revisit

nucleus/dataset.py
bdnorman: Synced live: going to specify local cache vs s3 cache

bdnorman (approved):

nucleus/dataset.py
```diff
@@ -1439,11 +1454,76 @@ def items_and_annotation_generator(
...
+        )
+        for item in json_generator:
+            image = fetch_image(item[ITEM_KEY][PROCESSED_URL_KEY])
+            w, h = image.size
+            annotations = item[BOX_TYPE]
+            item_ref_id = item[ITEM_KEY][REFERENCE_ID_KEY]
+            for x0, y0 in generate_offsets(w, h, chip_size, stride_size):
+                x0, y0 = int(x0), int(y0)
+                x1 = x0 + chip_size
+                y1 = y0 + chip_size
```
gatli, line 1506: Are you intentionally not doing any clamping to the image dimensions here? Not sure what PIL does under the hood here with out of bounds coords.

nucleus/utils.py
```diff
@@ -422,3 +435,120 @@ def paginate_generator(
...
+    s3_bucket, s3_key = s3_path.split("//", 1)[-1].split("/", 1)
+    return s3_bucket, s3_key
+
+
+def fetch_image(s3_url: str):
+    s3_bucket, s3_key = split_s3_bucket_key(s3_url)
+    image = Image.open(
+        boto3.resource("s3").Bucket(s3_bucket).Object(s3_key).get()["Body"]
+    )
+    return image
```
gatli, line 450: Let's split this up into a seperable s3 util module.

nucleus/utils.py
```diff
@@ -422,3 +435,120 @@ def paginate_generator(
...
+
+
+def fetch_chip(ref_id: str):
```
gatli, line 453: All of the chip utils are so specific that I'd put them in a separate chip_util file.

gatli, line 453: This needs a docstring, what do all of the permutations of the return values mean? I'd also consider composing the ref_id in this function rather than outside.

nucleus/utils.py
```diff
@@ -422,3 +435,120 @@ def paginate_generator(
...
+    xs = np.arange(0, w - stride_size, chip_size - stride_size)
+    ys = np.arange(0, h - stride_size, chip_size - stride_size)
+    if len(xs) > 1:
+        xs = np.round(xs * (w - chip_size) / xs[-1]).astype(int)
+    if len(ys) > 1:
+        ys = np.round(ys * (h - chip_size) / ys[-1]).astype(int)
+    yield from product(ys, xs)
+
+
+def chip_annotations(data, x0: int, y0: int, x1: int, y1: int):
```
gatli, line 526: This needs a docstring explaining what's going on. Why are you creating annotations for each chip?

nucleus/dataset.py
```diff
@@ -1439,11 +1454,76 @@ def items_and_annotation_generator(
...
                 yield ia
 
+    def items_and_annotation_chip_generator(
+        self,
+        chip_size: int,
+        stride_size: int,
+        cache_directory: str,
+        query: Optional[str] = None,
+    ) -> Iterable[Dict[str, str]]:
+        """Provides a generator of chips for all DatasetItems and BoxAnnotations in the dataset.
```
gatli, line 1470: I'd explain what a chip is here. The term I'm most familiar with is tiling.

nucleus/utils.py
```diff
@@ -422,3 +435,120 @@ def paginate_generator(
...
+                s3_key + ".json"
+            ).put(
+                Body=json.dumps(annotations, ensure_ascii=False).encode(
+                    "UTF-8"
+                ),
+                ContentType="application/json",
+            )
+            annotation_loc = ref_id + ".json"
+        return ref_id + ".jpeg", annotation_loc
+    else:
```
gatli, line 503: Is there no danger in treating everything except s3 as local?

vayunalapati: since we clarify in the doc strings we only support s3 and local users should expect an error if they use a different type

PR #397, opened 2023-10-03
Handle error JobStatus in sleep_until_complete

gatli:
The `AsyncJob.sleep_until_complete()` only handled the deprecated error status. This should fix the problems Covariant where facing.


ntamas92 (approved):
LGTM

PR #402, opened 2023-10-19
Implement dataset.query_objects method

ntamas92:


nucleus/dataset.py
```diff
@@ -98,6 +101,15 @@
 WARN_FOR_LARGE_SCENES_UPLOAD = 5
 
 
+# TODO: This might not be called objectType, but rather queryType?
+class ObjectType(str, Enum):
+    IOU = ("iou",)
+    FALSE_POSITIVE = ("false_positive",)
+    FALSE_NEGATIVE = ("false_negative",)
+    PREDICTIONS_ONLY = ("predictions_only",)
+    GROUND_TRUTH_ONLY = ("ground_truth_only",)
```
gatli, line 110: Why are these one-tuples instead of strings?

nucleus/iou_match.py
```diff
@@ -0,0 +1,75 @@
...
+    IOU_KEY,
+    MODEL_PREDICTION_ID_KEY,
+    MODEL_PREDICTION_LABEL_KEY,
+    MODEL_RUN_ID_KEY,
+)
+
+
+class ConfusionCategory(Enum):
+    TRUE_POSITIVE = "true_positive"
+    TRUE_NEGATIVE = "true_negative"
```
gatli, line 17: I think we can safely drop this for our computer vision use case 🙂

nucleus/iou_match.py
```diff
@@ -0,0 +1,75 @@
...
+    model_prediction_label: str,
+):
+    confusion_category = ConfusionCategory.FALSE_NEGATIVE
+
+    if (
+        model_prediction_label is None
+        and ground_truth_annotation_label is None
+    ):
+        confusion_category = ConfusionCategory.TRUE_NEGATIVE
+    elif model_prediction_label == ground_truth_annotation_label:
```
gatli, line 33: This is wrong as it doesn't take into account allowed label mapping. We can have a true positive according to this mapping even though the labels don't match.

The `true_positive` field indicates if the match was a match.

gatli, line 33:
```suggestion
    if model_prediction_label == ground_truth_annotation_label:
```
Never happens for our use cases. Let's just drop this.

nucleus/iou_match.py
```diff
@@ -0,0 +1,75 @@
...
+        confusion_category = ConfusionCategory.TRUE_POSITIVE
+    elif model_prediction_label is not None:
+        confusion_category = ConfusionCategory.FALSE_POSITIVE
+
+    return confusion_category
+
+
+# TODO: Is there a better name for this class?
+@dataclass
+class IOUMatch:
```
gatli, line 43: Yeah, there probably is. Maybe we won't always match by IOU but another distance metrics. 

I'll have to think about this a bit...

gatli: EvaluationMatch? It's distance metric agnostic...

nucleus/iou_match.py
```diff
@@ -0,0 +1,75 @@
...
+    @classmethod
+    def from_json(cls, payload: dict):
+        model_prediction_label = payload.get(MODEL_PREDICTION_LABEL_KEY, None)
+        ground_truth_annotation_label = payload.get(
+            GROUND_TRUTH_ANNOTATION_LABEL_KEY, None
+        )
+        confusion_category = infer_confusion_category(
+            ground_truth_annotation_label=ground_truth_annotation_label,
+            model_prediction_label=model_prediction_label,
+        )
```
gatli, line 62: Maybe this was a bit of a confusing ask: we should propogate the true_positive to the SDK but then use that to map the confusion type. 🙂

gatli (commented):
Looking good except some Nucleus specific logic 🙂 

You might have inferred a wrong class based on a mapping based match ...

nucleus/dataset.py
ntamas92: Haha, this is just my Python starter's issue. I've left a trailing comma at the end of the enums, and black has thought that I wanted to create a tuple ¯\_(ツ)_/¯

nucleus/iou_match.py
ntamas92: I'm not too familiar with the jargon yet, but I like it.

gatli (approved):
LGTM 👍 

I added a docstring for EvaluationMatch, LMK if there is anything unclear there.
- fixed IOU type `int`->`float`

Nullable has the lame type annotation `Optional`. You'll get used to it 😉

ntamas92:
> LGTM 👍
> 
> I added a docstring for EvaluationMatch, LMK if there is anything unclear there.
> 
> * fixed IOU type `int`->`float`
> 
> Nullable has the lame type annotation `Optional`. You'll get used to it 😉

Thanks! Yeah, originally I wrote `str | None` to mark nullable properties, but that seemed to break the builds prior to 3.10

PR #404, opened 2023-10-30
Make skeleton key optional on KeypointsAnnotation

ntamas92:


jean-lucas (approved):
nice, LGTM thanks for fixing it so fast

PR #405, opened 2023-10-30
Allow embedding info upload for Dataset Items

ntamas92:
This change allows embedding vector information to be directly attached to DatasetItems. This will enable utilizing Nucleus in on-premise use cases where the image information and embedding computation is not done on Scale side.

This change requires information about the search indexes (more specifically the ID of the indexes) for a dataset, which the SDK currently does not provide.


gatli (approved):
Nice LGTM 👍 

I guess next step would be to expose index_ids for privacy mode datasets?

nucleus/dataset_item.py
```diff
@@ -112,17 +127,25 @@ class DatasetItem:  # pylint: disable=R0902
     )
     metadata: Optional[dict] = None
     pointcloud_location: Optional[str] = None
+    # TODO: Should we disallow the embedding info upload if this is true?
     upload_to_scale: Optional[bool] = True
+    embedding_info: Optional[DatasetItemEmbeddingInfo] = None
```
jean-lucas, line 132: I had the same question in my privacy dataset PR.
Lets not worry about it for now.

jean-lucas (approved):
nice. Lets also update the changelog with a 1-3 line demo

pfmark:
> I guess next step would be to expose index_ids for privacy mode datasets?

Yes we need a dataset method to expose index ids of custom indexes through the SDK. @ntamas92 let's also add this in a fwup PR.

ntamas92:
> I guess next step would be to expose index_ids for privacy mode datasets?

@gatli yeah I think so. Should it be limited to `privacy_mode` datasets though?

gatli:
> > I guess next step would be to expose index_ids for privacy mode datasets?
> 
> @gatli yeah I think so. Should it be limited to `privacy_mode` datasets though?

I think it's fine to expose the index_ids for every repo but limit appending to privace_mode for CLIP indexes.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to the [Nucleus Python Client](https://github.com/scaleapi/n
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.16.7](https://github.com/scaleapi/nucleus-python-client/releases/tag/v0.16.7) - 2023-11-03
+
+### Added
+- Allow direct embedding vector upload together with dataset items. `DatasetItem` now has an additional parameter called `embeddingInfo` which can be used to directly upload embeddings when a dataset is uploaded.
```
gatli, line 11: snake 🐍 
```suggestion
- Allow direct embedding vector upload together with dataset items. `DatasetItem` now has an additional parameter called `embedding_info` which can be used to directly upload embeddings when a dataset is uploaded.
```

gatli (approved):
Alles gut 👍

PR #407, opened 2023-11-03
Remove upload_to_scale option

jean-lucas:
Users should now use the `use_privacy_mode` on the Dataset level instead.

I'll make a release for this later on, its not too critical as no one is using the upload_to_scale flag right now anyways


nucleus/dataset_item.py
```diff
@@ -127,24 +119,19 @@ class DatasetItem:  # pylint: disable=R0902
     )
     metadata: Optional[dict] = None
     pointcloud_location: Optional[str] = None
-    upload_to_scale: Optional[bool] = True
```
gatli, line 130: We sure that nobody is using this? Otherwise I'd keep this -> change it to None and add a warning if this is not None that you should create a privacy dataset.

gatli (approved):
LGTM 👍 

On the off chance that somebody is using this we could keep the parameter and throw an error with information about privacy mode.

jean-lucas:
> Did you want to create a changelog entry for this, or just do it once you do the release?

Yes, I'll do it on the next release

nucleus/dataset_item.py
jean-lucas: I checked all dataset items, no one uses this feature (it was also broken before too)

jean-lucas:
>On the off chance that somebody is using this

No one is 🙌

PR #409, opened 2023-11-09
add width/height for dataset item

jean-lucas:
- Allow passing width/height to dataset item
- Useful for privacy mode, where we cannot calculate width/height on our side


nucleus/dataset_item.py
```diff
@@ -237,3 +247,13 @@ def check_for_duplicate_reference_ids(dataset_items: Sequence[DatasetItem]):
...
+
+
+def check_items_have_dimensions(dataset_items: Sequence[DatasetItem]):
+    for item in dataset_items:
+        has_width = getattr(item, "width")
+        has_height = getattr(item, "height")
+        if not (has_width and has_height):
+            raise Exception(
+                f"When using privacy mode, all items require a width and height. Missing for item: '{item.reference_id}'"
+            )
```
jean-lucas, line 259: fyi @gatli your script for privacy data upload will break after this change.

But i believe you can already get the image dimensions from the embedding service anyways?

pyproject.toml
```diff
@@ -25,7 +25,7 @@ ignore = ["E501", "E741", "E731", "F401"]  # Easy ignore for getting it running
 
 [tool.poetry]
 name = "scale-nucleus"
-version = "0.16.7"
+version = "0.16.8"
```
jean-lucas, line 28:
```suggestion
version = "0.16.9"
```

nucleus/constants.py
```diff
@@ -45,6 +45,7 @@
 DATASET_NAME_KEY = "name"
 DATASET_PRIVACY_MODE_KEY = "use_privacy_mode"
 DATASET_SLICES_KEY = "slice_ids"
+DATASET_USE_PRIVACY_MODE = "use_privacy_mode"
```
gatli, line 48: I'm seeing double 🧑‍🦲 🧑‍🦲 . Four `"use_privacy_mode"` constants!

nucleus/dataset.py
```diff
@@ -184,6 +187,17 @@ def is_scene(self) -> bool:
...
+    def use_privacy_mode(self) -> bool:
+        """Whether or not the dataset was created for privacy mode."""
+        if self._use_privacy_mode is not None:
+            return self._use_privacy_mode
+        response = self._client.make_request(
+            {}, f"dataset/{self.id}/use_privacy_mode", requests.get
+        )[DATASET_USE_PRIVACY_MODE]
+        self._use_privacy_mode = response
+        return self._use_privacy_mode  # type: ignore
+
```
gatli, line 200: I'm a little vary about using round-trip checks for every append. I think this will introduce more flakiness in our pipelines given how often we're seeing 503's etc.

Ideally we'd return this for the dataset constructor  -> then we always have the data locally.
We should the very least cache this locally since this won't ever change.

nucleus/dataset_item.py
gatli: Thanks for the heads up 🙂

gatli (approved):
Can we `lru_cache` this or ideally just use the information passed for the constructor?

Otherwise LGTM 👍

nucleus/constants.py
jean-lucas: oops good catch

gatli: Where's my laugh? I make good joke! 😄

pyproject.toml
```diff
@@ -25,7 +25,7 @@ ignore = ["E501", "E741", "E731", "F401"]  # Easy ignore for getting it running
 
 [tool.poetry]
 name = "scale-nucleus"
-version = "0.16.7"
+version = "0.16.9"
```
ntamas92, line 28: Nitpick, as I see, version `0.16.8` has not been released so maybe it would make sense to combine these changes into that version instead of creating a new one?

ntamas92 (approved):

nucleus/constants.py
ntamas92: Here you go!

nucleus/dataset.py
jean-lucas: Good point, I changed `client.create_dataset` such that the properties `name`, `is_scene`, and `use_privacy_mode` are saved and dont require round-trip checks

jean-lucas:
> Can we `lru_cache` this or ideally just use the information passed for the constructor?

Opted to using the info passed during `client.create_dataset`.
Round trips are still necessary when using `get_dataset`, but we can fix that at a later point.

PR #410, opened 2023-11-17
fixes to support video privacy mode

jean-lucas:
- Privacy mode for video upload was broken
- This PR addresses these fixes


nucleus/dataset.py
```diff
@@ -795,6 +794,7 @@ def _append_video_scenes(
             )
 
         for scene in scenes:
+            scene.use_privacy_mode = self.use_privacy_mode
             scene.validate()
```
jean-lucas, line 798: adding this property from dataset -> scene simplifies validation later on, w/o having to pass parameters everywhere

PR #411, opened 2023-11-21
allow for concurrent task fetches

jean-lucas:
Making individual requests to `download_pointcloud_task(task_id, frame_num)` was slow.

A new method is proposed to allow concurrent fetches of multiple tasks, with `download_pointcloud_tasks(task_ids, frame_num)`

### DIFF
--- a/nucleus/async_utils.py
+++ b/nucleus/async_utils.py
@@ -1,7 +1,15 @@
 import asyncio
 import time
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, BinaryIO, Callable, Sequence, Tuple
+from typing import (
+    BinaryIO,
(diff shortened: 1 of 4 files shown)


nucleus/async_utils.py
```diff
@@ -165,36 +184,80 @@ async def _post_form_data(
...
+                    endpoint, session, response, sleep_time
+                )
+                if data is None:
+                    continue
+
+                progressbar.update(1)
+                return (request, data)
+
+
+async def _parse_async_response(endpoint, session, response, sleep_time):
```
jean-lucas, line 235: extracted from function above

gatli (approved):
LGTM 👍 🔥

PR #412, opened 2023-11-21
Create dataset from dir

jean-lucas:
Allow creating a dataset by crawling all images in a directory, recursively.

Also supports privacy mode datasets.

#### Example structure:
```
~/Documents/
    data/
        2022/
            - img01.png
            - img02.png
        2023/
            - img01.png
            - img02.png
```

#### Example Default:

```python
data_dir = "~/Documents/data"
client.create_dataset_from_dir(data_dir)
# this will create a dataset named "data" and will contain 4 images, with the ref IDs:
# ["2022/img01.png", "2022/img02.png", "2023/img01.png", "2023/img02.png"]
```

#### Example Privacy Mode:

This requires that a proxy (or file server) is setup  and can serve files _relative_ to the data_dir

```python
data_dir = "~/Documents/data"
client.create_dataset_from_dir(
    data_dir,
    dataset_name='my-dataset',
    use_privacy_mode=True,
    privacy_mode_proxy="http://localhost:5000/assets/"
)
```

This would create a dataset `my-dataset`, and when opened in Nucleus, the images would be requested to the path:
`<privacy_mode_proxy>/`, for example: `http://localhost:5000/assets/2022/img01.png`

### DIFF
--- a/nucleus/__init__.py
+++ b/nucleus/__init__.py
@@ -45,7 +45,8 @@ __all__ = [
 import datetime
 import os
 import warnings
-from typing import Any, Dict, List, Optional, Sequence, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
+
 
 import pydantic
 import requests
@@ -150,6 +151,7 @@ from .retry_strategy import RetryStrategy
 from .scene import Frame, LidarScene, VideoScene
 from .slice import Slice
 from .upload_response import UploadResponse
+from .utils import create_items_from_folder_crawl
 from .validate import Validate
 
 # pylint: disable=E1101
@@ -1177,3 +1179,42 @@ class NucleusClient:
             raise NoAPIKey()
 
         return api_key
+
+    def create_dataset_from_dir(
+        self,
+        dirname: str,
+        dataset_name: Optional[str] = None,
+        use_privacy_mode: bool = False,
+        file_globs: Tuple[str] = (".png", ".jpg", ".jpeg"),
+        skip_size_warning: bool = False,
+    ):
+        """
+        Create a dataset by recursively crawling through a directory.
+        A DatasetItem will be created for each unique image found.
+
+        Parameters:
+            dataset_name: If none is given, the parent folder name is used
+            use_privacy_mode: Whether the dataset should be treated as privacy
+            file_globs: Which image types should be considered
+            skip_size_warning: If False, it will throw an error if the script globs more than 500 images. This is a safety check in case the dirname has a typo, and grabs too much data.
+        """
+
+        # ensures path ends with a single slash
+        _dirname = os.path.join(os.path.expanduser(dirname), "")
+        if not os.path.exists(_dirname):
+            raise ValueError(
+                f"Given directory name: {dirname} does not exists. Searched in {_dirname}"
+            )
+
+        folder_name = os.path.basename(_dirname)
+        dataset_name = dataset_name or folder_name
+
+        dataset = self.create_dataset(
+            dataset_name, use_privacy_mode=use_privacy_mode
+        )
+
+        items = create_items_from_folder_crawl(
+            _dirname, file_globs, use_privacy_mode, skip_size_warning
+        )
+
+        dataset.append(items, asynchronous=False)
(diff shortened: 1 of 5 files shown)


nucleus/__init__.py
```diff
@@ -1177,3 +1179,52 @@ def _set_api_key(self, api_key):
...
+            _dirname, file_ext_globs, use_privacy_mode, privacy_mode_proxy, skip_size_warning
+        )
+        if len(items) == 0:
+            print(f"Did not find any items in {dirname}")
+            return
+
+        dataset = self.create_dataset(
+            name=dataset_name, use_privacy_mode=use_privacy_mode
+        )
+        dataset.append(items, asynchronous=False)
```
jean-lucas, line 1228: We can run it with `asynchronous=True` when `use_privacy_mode` is true, but this should be fine for now

nucleus/utils.py
```diff
@@ -422,3 +425,55 @@ def paginate_generator(
...
+        fpaths = glob.glob(pathname=pathname, root_dir=dirname, recursive=True)
+        relative_fpaths.extend(fpaths)
+
+        if (
+            len(relative_fpaths) > GLOB_SIZE_THRESHOLD_CHECK
+            and not skip_size_warning
+        ):
+            raise Exception(
+                f"Found over {GLOB_SIZE_THRESHOLD_CHECK} items in {dirname}. If this is intended, set skip_size_warning=True when calling this function."
+            )
```
jean-lucas, line 458: I can remove this, if you think its unnecessary.
Its just in case some one mis-types a folder name, and we end up globbing way more than we should.

nucleus/utils.py
```diff
@@ -422,3 +425,55 @@ def paginate_generator(
...
+
+def get_image_dimension(image_fpath: str) -> Tuple[int, int]:
+    im = Image.open(image_fpath)
+    return im.size
+
+
+GLOB_SIZE_THRESHOLD_CHECK = 500
+
+
+def create_items_from_folder_crawl(
```
ntamas92, line 438: Maybe split this method into two? One part would be responsible for globbing, the other could just expect a list of paths and handle the datasetitem creation from them. Maybe we can reuse the second one in other cases as well

nucleus/utils.py
ntamas92: I think this is a fairly reasonable check to be had, I'd maybe put it into the client method instead though, that way we wouldn't need to pass the `skip_size_warning` here either

nucleus/__init__.py
```diff
@@ -1177,3 +1179,52 @@ def _set_api_key(self, api_key):
...
 
         return api_key
+
+    def create_dataset_from_dir(
+        self,
+        dirname: str,
+        dataset_name: Optional[str] = None,
+        use_privacy_mode: bool = False,
+        privacy_mode_proxy: str = "",
+        file_ext_globs: Tuple[str] = ("png", "jpg", "jpeg"),
```
ntamas92, line 1189: I think I'd give a less technical name for this and hide the underlying implementation, after all, these are just file extensions, not globs. (Also I just encountered the term glob now for the first time. I know, it's embarrassing 🙈)

ntamas92 (approved):
Few nitpicks otherwise seems awesome!

nucleus/__init__.py
jean-lucas: Changed it to `allowed_file_types`

nucleus/utils.py
jean-lucas: >I'd maybe put it into the client method instead though

The benefit of having it here _inside_ the for-loop is that it will break as soon as we glob more than the threshold.
Otherwise we would have to wait for a large glob to complete, create items, and then check the threshold.

nucleus/utils.py
jean-lucas: A bit tricky to create the 2nd function as it has some custom logic for privacy mode.
But I refactored the first part into its own function now.

pfmark:
lgtm overall right now. However, this will be without local embedding computation. We need to add this in the near future too then.

jean-lucas:
> However, this will be without local embedding computation. We need to add this in the near future too then.

Shouldn't be too hard to modify this code to support embedding computations

gatli:
> This requires that a proxy (or file server) is setup and can serve files relative to the data_dir

Just saw this -> I think we should generally go with absolute paths. We then set up the proxy to do path mangling if it only has access to relative paths, this should be more future proof.

PR #413, opened 2023-11-21
Add Parallel Processing Input to Item Chip Generator

vayunalapati:
Allow users to specify the number of processes they would like to spin up for the image chip generator


gatli (approved):
LGTM 👍

PR #415, opened 2023-11-29
Support Pydantic v2 environments via falling back to v1 all of the time

gatli:
Pydantic is very commonly used in codebases for payload validation. Adding scale-nucleus as a dependency pins users to Pydantic v1 which is totally incompatible with Pydantic v2. 

We solve this problem by opening up the dependencies to support `>=v1` and making sure that we always pick the v1 layer that is included in v2.


jean-lucas (approved):
lgtm

PR #418, opened 2023-11-30
Align X Y for Image Chip Generator

vayunalapati:
There was a mismatch between the x and y ordering of coordinates between https://github.com/scaleapi/nucleus-python-client/blob/master/nucleus/chip_utils.py#L125 and https://github.com/scaleapi/nucleus-python-client/blob/master/nucleus/chip_utils.py#L184


gatli (approved):
LGTM 👍 🙈

PR #417, opened 2023-11-30
Allow Users to Specify Trained Slice for Model Runs

vayunalapati:
Allow users to specify the slice id which was used to train a model.

PR #419, opened 2023-12-20
Updated model to include trained_slice_id

schumry:
Added trained_slice_id field to the model object.

This was previously added to [upload_predictions](https://github.com/scaleapi/nucleus-python-client/blob/master/nucleus/dataset.py#L1797) in [Vayun's PR](https://github.com/scaleapi/nucleus-python-client/pull/417). I am still learning my way around the repo, so please let me know if this addition is correct or necessary.

### DIFF
--- a/nucleus/model.py
+++ b/nucleus/model.py
@@ -119,7 +119,7 @@ class Model:
         self.trained_slice_id = trained_slice_id
 
     def __repr__(self):
-        return f"Model(model_id='{self.id}', name='{self.name}', reference_id='{self.reference_id}', metadata={self.metadata}, bundle_name={self.bundle_name}, tags={self.tags}, client={self._client}, trained_slice_id={self.trained_slice_id})"
+        return f"Model(model_id='{self.id}', name='{self.name}', reference_id='{self.reference_id}', metadata={self.metadata}, bundle_name={self.bundle_name}, tags={self.tags}, client={self._client}"
 
     def __eq__(self, other):
         return (
--- a/nucleus/model.py
+++ b/nucleus/model.py
@@ -3,7 +3,13 @@ from typing import Dict, List, Optional, Union
 import requests
 
 from .async_job import AsyncJob
-from .constants import METADATA_KEY, MODEL_TAGS_KEY, NAME_KEY, REFERENCE_ID_KEY
+from .constants import (
+    METADATA_KEY,
+    MODEL_TAGS_KEY,
+    NAME_KEY,
+    REFERENCE_ID_KEY,
+    TRAINED_SLICE_ID_KEY,
+)
 from .dataset import Dataset
 from .model_run import ModelRun
 from .prediction import (
@@ -101,6 +107,7 @@ class Model:
         client,
         bundle_name=None,
         tags=None,
+        trained_slice_id=None,
     ):
         self.id = model_id
         self.name = name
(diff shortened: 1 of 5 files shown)


ntamas92:
What are you trying to achieve with this PR? Just exposing the `trained_slice_id` on the models? 

Let's untangle a few concepts as models and model runs are a confusing topic, and usually being referred to with the wrong names.
On the Nucleus side, there are two model concepts: **Model Projects** (or just **Models**) and **Model Runs**. Models are the blueprint of an ML model, and Model Runs represent actual executions of that model, tied to a single dataset. 

There were some efforts to deprecate Model Runs from the client, but that's currently half-baked, and model runs are still first class citizens in the Nucleus side. If you check the [NucleusModelRun](https://github.com/scaleapi/scaleapi/blob/master/server/src/models/nucleus/NucleusModelRun.ts#L50C22-L50C22) entity in nucleus, you can find that the `trained_slice_id` is stored on it. If you wanna expose it, I'd suggest returning it by the [modelRun/:modelRunId/info](https://github.com/scaleapi/scaleapi/blob/master/server/src/routes/v1/select.ts#L1079) endpoint on the api side, and put it on the [info method](https://github.com/scaleapi/nucleus-python-client/blob/8b7c380e445bba92b87c9dc4411969fa2076213a/nucleus/model_run.py#L62) of `model_run.py`

@gatli can correct me if I said anything stupid :)

nucleus/model.py
```diff
@@ -109,9 +116,10 @@ def __init__(
         self.bundle_name = bundle_name
         self.tags = tags if tags else []
         self._client = client
+        self.trained_slice_id = trained_slice_id
 
     def __repr__(self):
-        return f"Model(model_id='{self.id}', name='{self.name}', reference_id='{self.reference_id}', metadata={self.metadata}, bundle_name={self.bundle_name}, tags={self.tags}, client={self._client})"
+        return f"Model(model_id='{self.id}', name='{self.name}', reference_id='{self.reference_id}', metadata={self.metadata}, bundle_name={self.bundle_name}, tags={self.tags}, client={self._client}, trained_slice_id={self.trained_slice_id})"
```
gatli, line 122: I'd drop this from the repr. Not commonly used and it is already way too long.

gatli (commented):
@ntamas92 is right that we need to think about Models and model runs. The trained slice id is associated with the model run and should continue to be exposed so. 

I think its still fine to add the methods to the model creation workflow since that creates a single model run under the hood. When reading the info we should just associate it with the model runs -> otherwise we have to aggregate over all model runs of a model.

nucleus/model.py
schumry: The only issue with removing it is that it fails the repr comparison test.

bdnorman:
@ntamas92 @gatli sorry just to clarify here: how should someone go about associating a training slice id with their model run using the python sdk? Can it be called via `create_run`?

levinwil:
@bdnorman [this PR](https://github.com/scaleapi/nucleus-python-client/blob/master/nucleus/dataset.py#L1797https://github.com/scaleapi/nucleus-python-client/blob/master/nucleus/dataset.py%23L1797) enables specifying a training slice ID when uploading predictions. @ntamas92 @gatli is this currently the only way to associate a training slice ID with a model? And if so, is the intention that this remains the only way to do so?

PR #421, opened 2024-01-01
Updated model to include trained_slice_id

schumry:
This update is to add `trained_slice_id` to the model object. So that when we create a model (with `NUCLEUS_CLIENT.create_model`) we are able to associate a `trained_slice_id` with it.

I did not change the repr response, since it causes issues with the test comparison. Let me know if we want to simplify the repr and modify the test or leave as is.

Associated scaleapi PR: https://github.com/scaleapi/scaleapi/pull/74083
Previous PR: https://github.com/scaleapi/nucleus-python-client/pull/419


levinwil (approved):

edwardpark97:
Talked with Ryan on Slack, I think we'll sync up with @levinwil tomorrow to scope out exactly what he needs for his work. We can always backfill trained_slice_id in the PSQL database at a later date, so I don't think it's really necessary to push through ASAP

pfmark:
> Talked with Ryan on Slack, I think we'll sync up with @levinwil tomorrow to scope out exactly what he needs for his work. We can always backfill trained_slice_id in the PSQL database at a later date, so I don't think it's really necessary to push through ASAP

feel free to loop me in, happy to learn more about the context here

schumry:
@scaleapi/nucleus-eng @edwardpark97 PR is now ready for review with accompanying [scaleapi PR](https://github.com/scaleapi/scaleapi/pull/74083).

nucleus/dataset.py
```diff
@@ -438,13 +438,15 @@ def create_model_run(
         model_id: Optional[str] = None,
         metadata: Optional[Dict[str, Any]] = None,
         annotation_metadata_schema: Optional[Dict] = None,
+        trained_slice_ids: Optional[str] = None,
```
edwardpark97, line 441: does it still make sense to add this to `create_model_run` now that trained_slice_id is associated to the modelproject object?

schumry: This was not meant to be pushed, however, with it being added to model projects object I do not see when we would ever use it with model runs. So I guess that also begs the question of if we should revert the changes for adding it to model run or just leave it in case of future uses.

ntamas92: I'd go with reverting the changes which added it to model_run to eliminate the confusion

ntamas92 (approved):

nucleus/dataset.py
schumry: Just to clarify (since I reread what I sent and it was not super clear). Definitely remove it here, but I was referring to also reverting the changes in https://github.com/scaleapi/nucleus-python-client/pull/417 or should we leave those around?

ntamas92: Yes I was referring to reverting the changes in https://github.com/scaleapi/nucleus-python-client/pull/417, otherwise it would be really confusing for the user to decide which api to choose.

nucleus/payload_constructor.py
```diff
@@ -158,6 +162,7 @@ def construct_model_run_creation_payload(
     model_id: Optional[str],
     metadata: Optional[Dict],
     annotation_metadata_schema: Optional[Dict] = None,
+    trained_slice_ids: Optional[str] = None,
```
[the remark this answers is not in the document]
schumry: Yup, not sure why I added that here.

nucleus/dataset.py
schumry: Will create a separate PR to pull out #417 changes.

schumry:
@scaleapi/nucleus-eng If the PR looks good to y'all, would someone be able to merge it please.

gatli:
> @scaleapi/nucleus-eng If the PR looks good to y'all, would someone be able to merge it please.
@schumry I bumped the version and added a CHANGELOG - you need to bump the version for the new version to be deployed to PyPI. I'll merge this once tests pass.

PR #423, opened 2024-01-11
Concurrent lidar pcl fix

pfmark:
Before it would always overwrite the `LidarPoint` to `Point3D` which drops the intensity which is not desired. Test with any 3D task of an annotation project.


gatli (approved):
LGTM 👍

PR #428, opened 2024-02-01
fix: reverts test_dataset_update_metadata_local

nicolastomeo:
snake_field must be 1 since second dataset append overrides the existing one (update=True)


gatli (approved):
LGTM 👍

PR #430, opened 2024-02-05
feat: Adds update_dataset_from_dir

nicolastomeo:


nucleus/__init__.py
```diff
@@ -1285,15 +1296,88 @@ def create_dataset_from_dir(
 
         if len(items) == 0:
             print(f"Did not find any items in {dirname}")
-            return None
+            return existing_dataset
```
ntamas92, line 1299: I'm wondering if it would be actually better to create the dataset in this case also. Even though there were not items in the directory, we wanted to create the dataset in the first place.

nucleus/__init__.py
```diff
@@ -1285,15 +1296,88 @@ def create_dataset_from_dir(
...
+        return self._create_or_update_dataset_from_dir(
+            dirname,
+            dataset_name=dataset_name,
+            use_privacy_mode=use_privacy_mode,
+            privacy_mode_proxy=privacy_mode_proxy,
+            allowed_file_types=allowed_file_types,
+            skip_size_warning=skip_size_warning,
+        )
+
+    def update_dataset_from_dir(
```
ntamas92, line 1348: Maybe this one would be better to be had on the dataset class itself. I think it's a better workflow to create/fetch a dataset explicitly then ingesting items into it.

ntamas92 (commented):
Nice work!! Small comments regarding the usage of the functions.

nucleus/__init__.py
gatli: Agreed, I think that's better encapsulation

gatli (commented):
Looking good - can you also add a method to the CLI? See `cli/*` - this is a perfect use case for a CLI tool. LMK if you need an intro into `click` and `rich`.

nucleus/__init__.py
jean-lucas: That's a good point, i think the dataset should still be created. I think the original behaviour was misleading.

nucleus/__init__.py
jean-lucas: I think that's a good suggestion. It would avoid the user having to copy the dataset id.

I think we can even rename it to:

`dataset.add_images_from_dir(dirname)`

jean-lucas (approved):
overall LGTM, tried it out, worked well.
I agree with @ntamas92 comments.

Before merging though, lets update the changelog  and version

nucleus/__init__.py
nicolastomeo: Addressed

nicolastomeo:
> Looking good - can you also add a method to the CLI? See `cli/*` - this is a perfect use case for a CLI tool. LMK if you need an intro into `click` and `rich`.

Yes, but i think it's better to do it in a different PR

nucleus/__init__.py
nicolastomeo: Addressed

nucleus/dataset.py
```diff
@@ -2241,3 +2243,50 @@ def jobs(
...
+            existing_dirname = self._client.valid_dirname(dirname)
+        items = create_items_from_folder_crawl(
+            existing_dirname,
+            allowed_file_types,
+            self.use_privacy_mode,
+            privacy_mode_proxy,
+        )
+
+        if len(items) == 0:
+            print(f"Did not find any items in {dirname}.")
```
jean-lucas, line 2286:
```suggestion
            print(f"Did not find any items in {dirname}.")
            return;
```

jean-lucas: We can do an early return here to avoid initiating a job with no items. Since the dataset is already created anyways.

jean-lucas (approved):
LGTM 👍

PR #432, opened 2024-02-27
Add S3_ENDPOINT to customize S3 endpoint URL

edwardpark97:
In certain environments, the S3 Endpoint URL is nonstandard and needs to be set to a specific value. This is particularly important when asking the Scale backend for a presigned S3 URL.

This is in combination with this commit on the main Scale monorepo: https://github.com/scaleapi/scaleapi/pull/76269/commits/69a721081813249b7ecdd54bdaa18be60fb9d3ef


jean-lucas (approved):
LGTM once the change log conflicts are fixed

PR #433, opened 2024-02-27
Allow image+annotation generator to give mirrored URLs

edwardpark97:
We use the items_and_annotation_generator to fetch images and annotations for ML training. Sometimes the original image URL is no longer accessible to us and we need to fetch the Scale-mirrored URL instead.

I am open to changing the `use_mirrored_images` flag to whatever name is most memorable.

### DIFF
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -1447,6 +1447,7 @@ class Dataset:
     def items_and_annotation_generator(
         self,
         query: Optional[str] = None,
+        use_mirrored_images: bool = False,
     ) -> Iterable[Dict[str, Union[DatasetItem, Dict[str, List[Annotation]]]]]:
         """Provides a generator of all DatasetItems and Annotations in the dataset.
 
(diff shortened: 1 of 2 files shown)


levinwil (approved):

nucleus/dataset.py
```diff
@@ -1447,6 +1447,7 @@ def items_and_annotations(
     def items_and_annotation_generator(
         self,
         query: Optional[str] = None,
+        use_mirrored_images: bool = False,
```
jean-lucas, line 1450: Could you please add `use_mirrored_images` to the docstring of this method?

edwardpark97:
Docstring was updated, re-requesting review

Also heads up that I'll likely batch this PR alongside this PR https://github.com/scaleapi/nucleus-python-client/pull/432 and make them both part of 0.17.2 so the CHANGELOG will be updated there

jean-lucas (approved):
lgtm

PR #435, opened 2024-03-25
Add model run name for model.run

ntamas92:
Model run nam has been made required in scale api, this PR adds the `model_run_name` parameter to `Model.run`.

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -380,23 +380,6 @@ def test_annotate_async(dataset: Dataset):
     expected = {
         "job_id": job.job_id,
         "status": "Completed",
-        "message": {
-            "annotation_upload": {
-                "total": 4,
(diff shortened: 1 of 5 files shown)


tests/test_dataset.py
```diff
@@ -382,7 +382,6 @@ def test_annotate_async(dataset: Dataset):
         "status": "Completed",
         "message": {
```
gatli, line 383: This is a terrible testing pattern that we're still unwinding - we shouldn't be checking the message at all. We can test that in `scaleapi` unit tests if we want. 

Can change the test to use `assert_partial_equality` and ideally just check the `status` message or the subset that  indicates succes.

gatli (approved):
LGTM - I would simplify the test while we're at it 🙂

PR #439, opened 2024-07-03
Endpoint to fetch annotations grouped by scenes 

vinay553:
Exposes new endpoint added here: https://github.com/scaleapi/scaleapi/pull/88450 to the nucleus python client. Pretty bare bones, the format returned by the REST endpoint is what we want to expose to the customer.  

Will follow the steps to publish the updated library after this is approved.

### DIFF
--- a/nucleus/dataset.py
+++ b/nucleus/dataset.py
@@ -1453,6 +1453,7 @@ class Dataset:
         self,
         query: Optional[str] = None,
         use_mirrored_images: bool = False,
+        group_by_scene: bool = False,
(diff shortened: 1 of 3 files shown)


nucleus/dataset.py
```diff
@@ -1449,6 +1449,19 @@ def items_and_annotations(
         )
         return convert_export_payload(api_payload[EXPORTED_ROWS])
 
+    def scene_and_annotation_generator(self):
```
edwardpark97, line 1452: might want page_size as a parameter with a default size

edwardpark97 (approved):
please include a super detailed docstring as well, similar to all the other functions in the file

PR #444, opened 2024-12-07
[DE-4880] Update annotation related endpoints for multiple ground truth sets

iz18:
With [this feature](https://linear.app/scale-epd/issue/DE-4419), we will have multiple sets of ground truth if a dataset item is labeled across multiple tasks. To not suddenly break everything on the ML side, we want to have the option to return the legacy results, which is only the set of annotations that correspond to the most recent task. This is also changed to be consistent with the model results shown on Nucleus itself where it will only pull the most recent task. The task id is also now returned with all the annotations.

Moving forward, ML should keep track of which set of ground truth they trained their model with, and they can pull back that set of ground truth by setting `only_most_recent_task` to False. I think a new endpoint where they can request a set of ground truth based on a specific `task_id` would also be helpful in the future.

Accompanies [this pr](https://github.com/scaleapi/scaleapi/pull/100057)


daveguo-scale (approved):
tagging @scaleapi/field-eng-federal for visibility

iz18:
Removed the black formatting in circle ci because it doesn't behave the same way as local black formatter on the same version. Weird issue and maybe should be looked into? Will put a ticket into backlog.

.circleci/config.yml
```diff
@@ -34,10 +34,11 @@ jobs:
...
             include-python-in-cache-key: false
-      - run:
-          name: Black Formatting Check # Only validation, without re-formatting
-          command: |
-            poetry run black --check .
+      # - run:
+      #     name: Black Formatting Check # Only validation, without re-formatting
+      #     command: |
+      #       poetry show black
+      #       poetry run black --check .
```
jean-lucas, line 41: Please add this back.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to the [Nucleus Python Client](https://github.com/scaleapi/n
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.17.8](https://github.com/scaleapi/nucleus-python-client/releases/tag/v0.17.7) - 2024-11-05
+
```
jean-lucas, line 9: The date this wrong, and the URL is wrong, please update.

tests/test_prediction.py
```diff
@@ -58,25 +58,6 @@ def test_repr(test_object: any):
...
-    [
-        test_repr(CategoryPrediction.from_json(_))
-        for _ in TEST_CATEGORY_PREDICTIONS
-    ]
-
-    [
-        test_repr(CategoryPrediction.from_json(_))
-        for _ in TEST_DEFAULT_CATEGORY_PREDICTIONS
-    ]
-
```
jean-lucas, line 79: Why were these removed?

.circleci/config.yml
iz18: see thread: https://scaleapi.slack.com/archives/C05Q7DSPQF9/p1734546043282449, i think we need to up the black versioning, then python dependency version but not sure about how to still support testing for python versions less than 3.8

tests/test_prediction.py
iz18: removed them because the predictions were generated from the annotations, but now we have an extra field in the annotations so it generates predictions with an extra field and it breaks this test. i can change the test prediction generation and bring these tests back.

PR #446, opened 2025-03-11
[DE-5241] Add class label export

iz18:
Adding ability to extract class labels from a dataset or slice. Useful for MLEs using Nucleus to grab what objects are present in the dataset/slice they are working with.

Added two tests to check that the endpoint works.

Merge after https://github.com/scaleapi/scaleapi/pull/110796

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -309,9 +309,9 @@ def test_dataset_append_async(dataset: Dataset):
 
 def test_dataset_append_async_with_local_path(dataset: Dataset):
     ds_items = make_dataset_items()
-    ds_items[
-        0
-    ].image_location = "/a/fake/local/path/you/can/tell/is/local/but/is/fake"
+    ds_items[0].image_location = (
+        "/a/fake/local/path/you/can/tell/is/local/but/is/fake"
+    )
     with pytest.raises(ValueError):
         dataset.append(ds_items, asynchronous=True)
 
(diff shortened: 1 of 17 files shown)


iz18:
tests will fail until https://github.com/scaleapi/scaleapi/pull/110796 is merged

tests/test_dataset.py
```diff
@@ -610,3 +606,18 @@ def test_create_update_dataset_from_dir(CLIENT):
...
+@pytest.mark.integration
+def test_dataset_export_class_labels(dataset):
+    dataset.append(make_dataset_items())
+    # Create box annotation from the test data
+    box_annotation = BoxAnnotation(**TEST_BOX_ANNOTATIONS[0])
+    dataset.annotate(annotations=[box_annotation])
+
+    import time
+
+    time.sleep(60)
```
vinay553, line 620: I would just add a comment about this here (and try 30s)

tests/test_slice.py
```diff
@@ -96,17 +95,44 @@ def get_expected_item(reference_id):
...
+    slc = dataset.create_slice(
+        name=TEST_SLICE_NAME,
+        reference_ids=slice_ref_ids,
+    )
+
+    dataset.annotate(annotations=annotations)
+
+    import time
+
+    time.sleep(60)
```
vinay553, line 131: same here

vinay553 (approved):

PR #448, opened 2025-03-14
Add ungrouped function from deprecated class to model class

ankitvedakscale:
Reusing a function from the Deprecated ModelRun to Model Class

### DIFF
--- a/nucleus/model.py
+++ b/nucleus/model.py
@@ -4,6 +4,7 @@ import requests
 
 from .async_job import AsyncJob
 from .constants import (
+    ANNOTATIONS_KEY
     METADATA_KEY,
     MODEL_TAGS_KEY,
     MODEL_TRAINED_SLICE_IDS_KEY,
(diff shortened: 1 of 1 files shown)


nucleus/model.py
```diff
@@ -343,3 +345,10 @@ def remove_trained_slice_ids(self, slide_ids: List[str]):
             )
 
         return response.json()
+    def ungrouped_export(self, model_run_id):
+        json_response = self._client.make_request(
+            payload={},
+            route=f"modelRun/{model_run_id}/ungrouped",
+            requests_command=requests.get,
+        )
+        return format_prediction_response({ANNOTATIONS_KEY: json_response})
```
jean-lucas, line 354: Lets name this something a bit better. I'm not sure why the endpoint was originally named ungrouped, but this method could be called export_by_model_run_id or something like that

jean-lucas (commented):
Please also update the version in the .toml file and add a description in the releases

jean-lucas (approved):

PR #462, opened 2026-06-04
Add local pHash deduplication utility

vinay553:
## Summary

Adds a local pHash-based deduplication utility for Nucleus SDK users who already have a large list of `DatasetItem` objects or rows from `items_and_annotation_generator()`.

- Exposes `nucleus.deduplicate_by_phash(...)` and `LocalDeduplicationResult`.
- Accepts either `DatasetItem` objects or generator rows with an `"item"` `DatasetItem`.
- Preserves Nucleus threshold semantics for `0..64`.
- Uses fast exact-pHash handling for `threshold == 0`, a single-item fast path for `threshold == 64`, an exact chunked Hamming-neighbor index for `1 <= threshold <= 10`, and a linear exact fallback for thresholds above 10.
- Returns the surviving original objects, surviving dataset items, reference IDs, and `DeduplicationStats`.

### Baseline Problem

The logical algorithm is greedy deduplication. Sort rows deterministically, then keep a candidate only if no previously kept pHash is within the requested Hamming distance threshold.

```python
kept = []
for row in sorted_rows:
    if not any(hamming(row.phash, kept_hash) <= threshold for kept_hash in kept):
        kept.append(row)
```

This is exact but effectively quadratic for mostly distinct inputs, because each new candidate may scan every pHash kept so far.

### Technique 1: Fail Fast and Normalize Once

Before the dedup pass starts, each input is normalized into an internal record with the original object, its `DatasetItem`, a stable tie-break ID, and the pHash parsed as a 64-bit integer. Threshold validation and pHash validation happen up front, so a long local run fails before doing expensive work if any pHash is missing or malformed.

### Technique 2: Integer pHashes and Integer Hamming Distance

Backend pHashes arrive as 64-character binary strings. The utility parses each pHash once with `int(phash, 2)`, then computes Hamming distance with:

```python
(candidate_phash_value ^ kept_phash_value).bit_count()
```

This avoids per-character string comparison and enables sorting by `(phash_value, stable_id)`.

### Technique 3: Threshold-Specific Fast Paths

| Threshold case | Implementation | Reason |
|---|---|---|
| `threshold == 0` | Keep the first row per exact integer pHash. | Hamming distance must be exactly zero, so only exact pHash duplicates are removed. |
| `threshold == 64` | Keep the first sorted item. | Any two 64-bit hashes are within distance 64. |
| `1 <= threshold <= 10` | Use the exact chunked Hamming index. | This is the expected practical range and keeps chunk radius small. |
| `threshold > 10` | Use a linear greedy scan. | Larger radii make chunk neighborhoods dense enough that the index becomes less attractive. |

### Technique 4: Chunking With the Pigeonhole Principle

The accelerated path splits each 64-bit pHash into six chunks:

```python
INDEX_CHUNK_BITS = (11, 11, 11, 11, 10, 10)
```

For threshold `t`, the per-chunk radius is `floor(t / 6)`. If two pHashes have total Hamming distance at most `t`, at least one of the six chunks must be within that per-chunk radius. Otherwise every chunk would exceed the radius, forcing the total Hamming distance above `t`.

For `threshold=10`, the per-chunk radius is `1`. One partition performs `4 * 12 + 2 * 11 = 70` dictionary lookups: each 11-bit chunk checks the exact value plus 11 one-bit variants, and each 10-bit chunk checks the exact value plus 10 one-bit variants.

### Technique 5: Two Partitions With Cheap Rotation

A single chunk partition can produce many false candidates. The utility uses two exact partitions:

- Partition 0 splits the original 64-bit pHash into six contiguous chunks.
- Partition 1 rotates the pHash right by 8 bits, then splits into the same chunk widths.

The rotation is cheap:

```python
((phash_value >> 8) | (phash_value << 56)) & ((1 << 64) - 1)
```

The pigeonhole argument applies independently to both partitions, so true duplicates cannot be missed. A candidate must be found by both partitions before the exact 64-bit Hamming check runs.

### Technique 6: Reusable Candidate Marks Instead of Set Intersections

The permanent state remains simple: `kept_hashes` stores pHashes that survived deduplication, and chunk dictionaries point chunk values back to positions in `kept_hashes`. Those positions are called kept indexes.

Conceptually, each query intersects two candidate sets:

```python
partition_0_candidates = {17, 31, 44, 91, ...}
partition_1_candidates = {3, 17, 88, 91, ...}
intersection = {17, 91, ...}
```

The first implementation literally allocated Python `set()` objects and intersected them per candidate. The current implementation uses a reusable `bytearray` with one mark byte per kept pHash:

```python
# Partition 0
candidate_marks[17] = 1
candidate_marks[31] = 1
candidate_marks[44] = 1
candidate_marks[91] = 1

# Partition 1
if candidate_marks[17] == 1: exact_check(17)
if candidate_marks[88] == 1: exact_check(88)  # skipped; mark is zero
if candidate_marks[91] == 1: exact_check(91)
```

After each query, only touched indexes are cleared. This computes the same intersection while avoiding large temporary set allocations.

## Real Dataset Profile

I profiled this on an internal real production dataset with `278,587` rows. Because this repository is public, the dataset identifier and name are intentionally omitted from the PR body. The API key was not included in the report or the PR.

The profiler exported rows with `items_and_annotation_generator()`, held the exported rows in memory, and then timed only `deduplicate_by_phash(...)` so export/API time and local dedup CPU time are separated.

| Input rows | Export time | Missing pHashes | Malformed pHashes | Duplicate reference IDs |
|---:|---:|---:|---:|---:|
| 278,587 | 127.04s | 0 | 0 | 0 |

Local dedup timing only:

| Threshold | Kept | Removed | Removed % | Dedup time | Peak tracemalloc | Process max RSS |
|---:|---:|---:|---:|---:|---:|---:|
| 0 | 277,164 | 1,423 | 0.51% | 1.86s | 47.90 MB | 1.41 GB |
| 5 | 252,084 | 26,503 | 9.51% | 50.79s | 66.94 MB | 1.46 GB |
| 10 | 194,776 | 83,811 | 30.08% | 396.58s | 59.02 MB | 1.46 GB |

The practical read: the expensive exact local dedup case took about `6m 36.6s` for `278,587` real rows. If a user exports the dataset and runs threshold 10, the observed end-to-end time for this dataset would be roughly `127.04s + 396.58s = 523.62s`, or about `8m 43.6s`.

That is a reasonable runtime for an offline local utility, especially because the implementation is deterministic, exact, requires no server-side job, and validates pHashes before starting the expensive pass.

## Validation

- `NUCLEUS_PYTEST_API_KEY=dummy .venv/bin/python -m pytest tests/test_local_deduplication.py -q`
- Real dataset profile using the local SDK editable install in `.venv`.

Focused tests cover row preservation, `DatasetItem` inputs, threshold 0, threshold 64, invalid thresholds, malformed pHashes, unsupported input shapes, custom sort keys, and parity against a simple linear baseline for thresholds 1, 5, and 10.

<h3>Greptile Summary</h3>

This PR adds a fully offline `deduplicate_by_phash()` utility and `LocalDeduplicationResult` to the Nucleus SDK. It requires no API key, accepts `DatasetItem` objects or generator rows, and dispatches to an exact-match path (threshold 0), a two-partition chunked Hamming index (thresholds 1–10), a linear greedy fallback (thresholds 11–63), or a keep-one path (threshold 64).

- **`nucleus/local_deduplication.py`**: New module implementing the full pipeline — pHash validation, stable-id tie-breaking via `_tie_break_key`, the `_HammingIndex` with pigeonhole-based candidate filtering and reusable `bytearray` marks, and the three dedup paths.
- **`tests/test_local_deduplication.py`**: Covers all dispatch paths, custom/integer/None sort keys, ordinal fallback, missing reference IDs, and parity against a linear baseline for thresholds 1, 5, 10, 11, and 20.
- **`nucleus/__init__.py`**: Re-exports `deduplicate_by_phash` and `LocalDeduplicationResult` from the public namespace.

Safe to merge — the new utility is fully offline and additive, with no changes to existing API paths or data models.

The algorithm is provably correct: the pigeonhole argument holds for both partitions independently, the mark-based candidate intersection correctly resets state after every query, and all three dedup paths produce results consistent with the linear baseline. The previous issues around string-encoded sort keys and empty-string reference IDs are both resolved. The type annotation for sort_key's return type is the only gap.

No files require special attention.
</details>

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| nucleus/local_deduplication.py | New file: implements the full local pHash dedup pipeline — exact path, two-partition Hamming index, and linear fallback — with correct mark-based candidate intersection and clean threshold dispatch. |
| tests/test_local_deduplication.py | New file: comprehensive test suite covering all fast-paths, sort-key variants, error cases, and a parity baseline for thresholds 1, 5, 10, 11, and 20 (index and linear-scan paths). |
| nucleus/__init__.py | Exports LocalDeduplicationResult and deduplicate_by_phash from the public nucleus namespace. |

</details>

```mermaid
flowchart TD
    A[deduplicate_by_phash] --> B[_validate_threshold]
    B --> C[_normalize_records]
    C --> D{empty?}
    D -- yes --> E[return empty result]
    D -- no --> F[sort by phash_value, stable_id]
    F --> G{threshold?}
    G -- 0 --> H[_deduplicate_exact]
    G -- 64 --> I[keep records 0]
    G -- 1..10 --> J[_deduplicate_with_index]
    G -- 11..63 --> K[_deduplicate_with_linear_scan]
    H --> L[LocalDeduplicationResult]
    I --> L
    J --> L
    K --> L
```
</details>

<a href="https://app.greptile.com/api/ide/cursor?prompt=Fix%20the%20following%201%20code%20review%20issue.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0Anucleus%2Flocal_deduplication.py%3A162%0AThe%20%60sort_key%60%20callable's%20return%20type%20is%20annotated%20as%20%60Union%5Bstr%2C%20int%5D%60%2C%20but%20%60_tie_break_key%60%20explicitly%20handles%20a%20%60None%60%20return%2C%20and%20%60test_deduplicate_by_phash_preserves_ordinal_fallback_order%60%20relies%20on%20it%20with%20%60sort_key%3Dlambda%20row%3A%20None%60.%20A%20caller%20reading%20the%20signature%20has%20no%20indication%20that%20returning%20%60None%60%20is%20valid%20and%20falls%20back%20to%20ordinal%20ordering%20%E2%80%94%20the%20annotation%20should%20reflect%20this.%0A%0A%60%60%60suggestion%0A%20%20%20%20sort_key%3A%20Optional%5BCallable%5B%5BInputT%5D%2C%20Optional%5BUnion%5Bstr%2C%20int%5D%5D%5D%5D%20%3D%20None%2C%0A%60%60%60%0A%0A&pr=462&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCursorDark.svg?v=3"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCursor.svg?v=3">Fix All in Cursor</picture></a> <a href="https://app.greptile.com/ide/claude-code?prompt=Fix%20the%20following%201%20code%20review%20issue.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0Anucleus%2Flocal_deduplication.py%3A162%0AThe%20%60sort_key%60%20callable's%20return%20type%20is%20annotated%20as%20%60Union%5Bstr%2C%20int%5D%60%2C%20but%20%60_tie_break_key%60%20explicitly%20handles%20a%20%60None%60%20return%2C%20and%20%60test_deduplicate_by_phash_preserves_ordinal_fallback_order%60%20relies%20on%20it%20with%20%60sort_key%3Dlambda%20row%3A%20None%60.%20A%20caller%20reading%20the%20signature%20has%20no%20indication%20that%20returning%20%60None%60%20is%20valid%20and%20falls%20back%20to%20ordinal%20ordering%20%E2%80%94%20the%20annotation%20should%20reflect%20this.%0A%0A%60%60%60suggestion%0A%20%20%20%20sort_key%3A%20Optional%5BCallable%5B%5BInputT%5D%2C%20Optional%5BUnion%5Bstr%2C%20int%5D%5D%5D%5D%20%3D%20None%2C%0A%60%60%60%0A%0A&repo=scaleapi%2Fnucleus-python-client&pr=462&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=3"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=3">Fix All in Claude Code</picture></a> <a href="https://app.greptile.com/api/ide/codex?prompt=IMPORTANT%3A%20Work%20in%20the%20repository%20%22scaleapi%2Fnucleus-python-client%22%20on%20the%20existing%20branch%20%22codex%2Flocal-phash-dedup%22.%20Checkout%20that%20branch%20%E2%80%94%20do%20NOT%20create%20a%20new%20branch%20or%20open%20a%20new%20PR.%20Push%20your%20changes%20to%20%22codex%2Flocal-phash-dedup%22.%0A%0AFix%20the%20following%201%20code%20review%20issue.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0Anucleus%2Flocal_deduplication.py%3A162%0AThe%20%60sort_key%60%20callable's%20return%20type%20is%20annotated%20as%20%60Union%5Bstr%2C%20int%5D%60%2C%20but%20%60_tie_break_key%60%20explicitly%20handles%20a%20%60None%60%20return%2C%20and%20%60test_deduplicate_by_phash_preserves_ordinal_fallback_order%60%20relies%20on%20it%20with%20%60sort_key%3Dlambda%20row%3A%20None%60.%20A%20caller%20reading%20the%20signature%20has%20no%20indication%20that%20returning%20%60None%60%20is%20valid%20and%20falls%20back%20to%20ordinal%20ordering%20%E2%80%94%20the%20annotation%20should%20reflect%20this.%0A%0A%60%60%60suggestion%0A%20%20%20%20sort_key%3A%20Optional%5BCallable%5B%5BInputT%5D%2C%20Optional%5BUnion%5Bstr%2C%20int%5D%5D%5D%5D%20%3D%20None%2C%0A%60%60%60%0A%0A&repo=scaleapi%2Fnucleus-python-client&pr=462&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodexDark.svg?v=3"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodex.svg?v=3">Fix All in Codex</picture></a>

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
Fix the following 1 code review issue. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 1
nucleus/local_deduplication.py:162
The `sort_key` callable's return type is annotated as `Union[str, int]`, but `_tie_break_key` explicitly handles a `None` return, and `test_deduplicate_by_phash_preserves_ordinal_fallback_order` relies on it with `sort_key=lambda row: None`. A caller reading the signature has no indication that returning `None` is valid and falls back to ordinal ordering — the annotation should reflect this.

```suggestion
    sort_key: Optional[Callable[[InputT], Optional[Union[str, int]]]] = None,
```

`````

</details>

<sub>Reviews (8): Last reviewed commit: ["Run pre-commit formatting"](https://github.com/scaleapi/nucleus-python-client/commit/a0409c02767e13f2a5d8a41a25e381c7089585e7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35648218)</sub>

### DIFF
--- a/nucleus/local_deduplication.py
+++ b/nucleus/local_deduplication.py
@@ -76,7 +76,7 @@ class _HammingIndex:
         self._hashes: List[int] = []
         self._candidate_marks = bytearray()
         self._indexes = [
-            [dict() for _ in range(INDEX_CHUNK_COUNT)]
+            [{} for _ in range(INDEX_CHUNK_COUNT)]
             for _ in range(PARTITION_COUNT)
         ]
 
@@ -94,9 +94,19 @@ class _HammingIndex:
         if not self._hashes:
             return None
 
+        touched_indexes = self._mark_partition_zero_candidates(phash_value)
+        if not touched_indexes:
+            return None
+
+        duplicate_index = self._find_marked_partition_one_duplicate(
+            phash_value
+        )
+        _clear_candidate_marks(self._candidate_marks, touched_indexes)
+        return duplicate_index
+
+    def _mark_partition_zero_candidates(self, phash_value: int) -> List[int]:
         touched_indexes: List[int] = []
         marks = self._candidate_marks
-
         partition_zero_chunks = _partition_chunks(phash_value, 0)
         for chunk_index, chunk_value in enumerate(partition_zero_chunks):
             index = self._indexes[0][chunk_index]
@@ -108,10 +118,12 @@ class _HammingIndex:
                     if marks[kept_index] == 0:
                         marks[kept_index] = 1
                         touched_indexes.append(kept_index)
+        return touched_indexes
 
-        if not touched_indexes:
-            return None
-
+    def _find_marked_partition_one_duplicate(
+        self, phash_value: int
+    ) -> Optional[int]:
+        marks = self._candidate_marks
         partition_one_chunks = _partition_chunks(phash_value, 1)
         for chunk_index, chunk_value in enumerate(partition_one_chunks):
             index = self._indexes[1][chunk_index]
@@ -125,11 +137,8 @@ class _HammingIndex:
                     if (
                         phash_value ^ self._hashes[kept_index]
                     ).bit_count() <= self._threshold:
-                        _clear_candidate_marks(marks, touched_indexes)
                         return kept_index
                     marks[kept_index] = 2
-
-        _clear_candidate_marks(marks, touched_indexes)
         return None
 
 
--- a/nucleus/local_deduplication.py
+++ b/nucleus/local_deduplication.py
@@ -75,12 +75,12 @@ class _HammingIndex:
         ]
         self._hashes: List[int] = []
         self._candidate_marks = bytearray()
-        self._indexes = [
+        self._indexes: List[List[dict[int, List[int]]]] = [
             [{} for _ in range(INDEX_CHUNK_COUNT)]
             for _ in range(PARTITION_COUNT)
         ]
 
-    def add(self, phash_value: int) -> None:
+    def _add(self, phash_value: int) -> None:
         kept_index = len(self._hashes)
         self._hashes.append(phash_value)
         self._candidate_marks.append(0)
@@ -90,7 +90,13 @@ class _HammingIndex:
                 index = self._indexes[partition_index][chunk_index]
                 index.setdefault(chunk_value, []).append(kept_index)
 
-    def find_duplicate_index(self, phash_value: int) -> Optional[int]:
+    def add_if_unique(self, phash_value: int) -> bool:
+        if self._find_duplicate_index(phash_value) is not None:
+            return False
+        self._add(phash_value)
+        return True
+
+    def _find_duplicate_index(self, phash_value: int) -> Optional[int]:
         if not self._hashes:
             return None
 
@@ -176,8 +182,6 @@ def deduplicate_by_phash(
 
     _validate_threshold(threshold)
     records = _normalize_records(objects, sort_key=sort_key)
-    records.sort(key=lambda record: (record.phash_value, record.stable_id))
-
     if not records:
         return LocalDeduplicationResult(
             unique=[],
@@ -189,6 +193,8 @@ def deduplicate_by_phash(
                 deduplicated_count=0,
             ),
         )
+
+    records.sort(key=lambda record: (record.phash_value, record.stable_id))
 
     if threshold == 0:
         unique_records = _deduplicate_exact(records)
@@ -305,17 +311,15 @@ def _deduplicate_with_index(
     index = _HammingIndex(threshold)
     unique_records = []
     for record in records:
-        if index.find_duplicate_index(record.phash_value) is not None:
-            continue
-        index.add(record.phash_value)
-        unique_records.append(record)
+        if index.add_if_unique(record.phash_value):
+            unique_records.append(record)
     return unique_records
 
 
 def _deduplicate_with_linear_scan(
     records: Sequence[_DeduplicationRecord[InputT]], threshold: int
 ) -> List[_DeduplicationRecord[InputT]]:
-    kept_hashes = []
+    kept_hashes: List[int] = []
     unique_records = []
     for record in records:
         is_duplicate = any(
--- a/nucleus/local_deduplication.py
+++ b/nucleus/local_deduplication.py
@@ -13,6 +13,7 @@ from .dataset_item import DatasetItem
 from .deduplication import DeduplicationStats
 
 InputT = TypeVar("InputT")
+_TieBreakKey = tuple[int, Union[int, str]]
 
 PHASH_REGEX = re.compile(r"^[01]{64}$")
 DEDUP_THRESHOLD_MIN = 0
@@ -36,18 +37,19 @@ class LocalDeduplicationResult(Generic[InputT]):
             ``DatasetItem`` objects.
         unique_dataset_items: The DatasetItem for each object in ``unique``.
         unique_reference_ids: Reference IDs for the DatasetItems in ``unique``.
+            Entries can be ``None`` if a DatasetItem has no reference ID.
         stats: Summary statistics for the run.
     """
 
     unique: List[InputT]
     unique_dataset_items: List[DatasetItem]
-    unique_reference_ids: List[str]
+    unique_reference_ids: List[Optional[str]]
     stats: DeduplicationStats
 
 
 @dataclass(slots=True)
 class _DeduplicationRecord(Generic[InputT]):
-    stable_id: str
+    stable_id: _TieBreakKey
     phash_value: int
     obj: InputT
     dataset_item: DatasetItem
@@ -194,7 +196,7 @@ def deduplicate_by_phash(
             record.dataset_item for record in unique_records
         ],
         unique_reference_ids=[
-            record.dataset_item.reference_id or "" for record in unique_records
+            record.dataset_item.reference_id for record in unique_records
         ],
         stats=DeduplicationStats(
             threshold=threshold,
@@ -238,9 +240,7 @@ def _normalize_records(
             if sort_key is not None
             else dataset_item.reference_id
         )
-        stable_id = (
-            str(tie_breaker) if tie_breaker is not None else str(ordinal)
-        )
+        stable_id = _tie_break_key(tie_breaker, ordinal)
         records.append(
             _DeduplicationRecord(
                 stable_id=stable_id,
@@ -250,6 +250,16 @@ def _normalize_records(
             )
         )
     return records
+
+
+def _tie_break_key(
+    tie_breaker: Optional[Union[str, int]], ordinal: int
+) -> _TieBreakKey:
+    if tie_breaker is None:
+        return (2, ordinal)
+    if isinstance(tie_breaker, int):
+        return (0, tie_breaker)
+    return (1, tie_breaker)
 
 
 def _extract_dataset_item(obj: InputT) -> DatasetItem:
--- a/nucleus/local_deduplication.py
+++ b/nucleus/local_deduplication.py
@@ -64,6 +64,13 @@ class _HammingIndex:
     at least one chunk within ``floor(t / 6)``. Querying both partitions and
     intersecting the candidates keeps the result exact while avoiding a full
     scan for the intended ``t <= 10`` case.
+
+    The chunk tolerance and the rotation are separate mechanisms. For
+    threshold 10, ``floor(10 / 6) == 1``, so a query chunk ``x`` probes the
+    exact bucket ``x`` plus the one-bit-away buckets ``x ^ (1 << bit_offset)``
+    inside that chunk. The eight-bit rotation does not implement that
+    tolerance; it gives a second chunk-boundary alignment so fewer false
+    candidates survive to the exact full-hash check.
     """
 
     def __init__(self, threshold: int):
--- a/nucleus/local_deduplication.py
+++ b/nucleus/local_deduplication.py
@@ -1,0 +1,349 @@
+"""Local pHash-based deduplication utilities."""
+
+from __future__ import annotations
+
+import itertools
+import re
+from collections.abc import Callable, Iterable, Mapping
+from dataclasses import dataclass
+from typing import Generic, List, Optional, Sequence, TypeVar, Union
+
+from .constants import ITEM_KEY
+from .dataset_item import DatasetItem
+from .deduplication import DeduplicationStats
+
+InputT = TypeVar("InputT")
+
+PHASH_REGEX = re.compile(r"^[01]{64}$")
+DEDUP_THRESHOLD_MIN = 0
+DEDUP_THRESHOLD_MAX = 64
+LOCAL_INDEX_MAX_THRESHOLD = 10
+PARTITION_COUNT = 2
+INDEX_CHUNK_BITS = (11, 11, 11, 11, 10, 10)
+INDEX_CHUNK_COUNT = len(INDEX_CHUNK_BITS)
+PHASH_VALUE_MASK = (1 << 64) - 1
+ROTATED_PARTITION_BITS = 8
+
+
+@dataclass
+class LocalDeduplicationResult(Generic[InputT]):
+    """Output of a local pHash deduplication run.
+
+    Attributes:
+        unique: Input objects that survived deduplication. If you passed rows
+            from ``items_and_annotation_generator()``, this contains those same
+            row dictionaries. If you passed ``DatasetItem`` objects, it contains
+            ``DatasetItem`` objects.
+        unique_dataset_items: The DatasetItem for each object in ``unique``.
+        unique_reference_ids: Reference IDs for the DatasetItems in ``unique``.
+        stats: Summary statistics for the run.
+    """
+
+    unique: List[InputT]
+    unique_dataset_items: List[DatasetItem]
+    unique_reference_ids: List[str]
+    stats: DeduplicationStats
+
+
+@dataclass(slots=True)
+class _DeduplicationRecord(Generic[InputT]):
+    stable_id: str
+    phash_value: int
+    obj: InputT
+    dataset_item: DatasetItem
+
+
+class _HammingIndex:
+    """Exact Hamming near-neighbor index for 64-bit hashes.
+
+    The first partition splits the pHash into six contiguous chunks.
+    The second partition rotates the pHash by eight bits, then applies the same
+    chunking. If two hashes are within threshold ``t``, then each partition has
+    at least one chunk within ``floor(t / 6)``. Querying both partitions and
+    intersecting the candidates keeps the result exact while avoiding a full
+    scan for the intended ``t <= 10`` case.
+    """
+
+    def __init__(self, threshold: int):
+        self._threshold = threshold
+        self._chunk_radius = threshold // INDEX_CHUNK_COUNT
+        self._variant_masks_by_chunk = [
+            _variant_masks(self._chunk_radius, chunk_bits)
+            for chunk_bits in INDEX_CHUNK_BITS
+        ]
+        self._hashes: List[int] = []
+        self._candidate_marks = bytearray()
+        self._indexes = [
+            [dict() for _ in range(INDEX_CHUNK_COUNT)]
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by vinay553** [comment]
@greptile-apps review this pr

**Comment by edwinpav** [comment]
@greptileai review

File: nucleus/local_deduplication.py [hunk 1]
```diff
@@ -0,0 +1,368 @@
...
+        ],
+        stats=DeduplicationStats(
+            threshold=threshold,
+            original_count=len(records),
+            deduplicated_count=len(unique_records),
+        ),
+    )
+
+
+def _validate_threshold(threshold: int) -> None:
```
edwinpav, line 231: is this validation not also happening in the original dedup sdk path? either way, can it be shared between both?

nucleus/local_deduplication.py
```diff
@@ -0,0 +1,368 @@
...
+    *,
+    sort_key: Optional[Callable[[InputT], Union[str, int]]],
+) -> List[_DeduplicationRecord[InputT]]:
+    records = []
+    for ordinal, obj in enumerate(objects):
+        dataset_item = _extract_dataset_item(obj)
+        phash = dataset_item.phash
+        if phash is None or not PHASH_REGEX.fullmatch(phash):
+            ref_id = dataset_item.reference_id
+            ref_msg = f" with reference_id {ref_id!r}" if ref_id else ""
```
edwinpav, line 254: nit: would it be useful to include the `dataset_item_id` as well?

nucleus/local_deduplication.py
```diff
@@ -0,0 +1,368 @@
...
+        return LocalDeduplicationResult(
+            unique=[],
+            unique_dataset_items=[],
+            unique_reference_ids=[],
+            stats=DeduplicationStats(
+                threshold=threshold,
+                original_count=0,
+                deduplicated_count=0,
+            ),
+        )
```
edwinpav, line 202: nbd but might be slightly more efficient to place this above the sort on 179? if i understand correctly the sort won't drop any records

nucleus/local_deduplication.py
```diff
@@ -0,0 +1,368 @@
...
+            for chunk_bits in INDEX_CHUNK_BITS
+        ]
+        self._hashes: List[int] = []
+        self._candidate_marks = bytearray()
+        self._indexes = [
+            [{} for _ in range(INDEX_CHUNK_COUNT)]
+            for _ in range(PARTITION_COUNT)
+        ]
+
+    def add(self, phash_value: int) -> None:
```
edwinpav: nit: should this also be `_add`

nucleus/local_deduplication.py
```diff
@@ -0,0 +1,368 @@
...
+        kept_index = len(self._hashes)
+        self._hashes.append(phash_value)
+        self._candidate_marks.append(0)
+        for partition_index in range(PARTITION_COUNT):
+            chunks = _partition_chunks(phash_value, partition_index)
+            for chunk_index, chunk_value in enumerate(chunks):
+                index = self._indexes[partition_index][chunk_index]
+                index.setdefault(chunk_value, []).append(kept_index)
+
+    def find_duplicate_index(self, phash_value: int) -> Optional[int]:
```
edwinpav: same as https://github.com/scaleapi/nucleus-python-client/pull/462#discussion_r3358603667

edwinpav (commented):
(meant to click approve)

Wowza what an optimization. Took me a while to understand the logic behind it. Only left some small nit comments. 

I'd suggest adding a comment like my explanation below within the code to help future readers.

Let me know if my understanding is correct (had ai format the wording):

We can safely skip most images without comparing against them, thanks to the pigeonhole principle. Chunks are defined by bit position, so we always compare a new pHash against a kept one chunk-by-chunk at the same positions — chunk 0 against chunk 0, chunk 1 against chunk 1, and so on. The principle applies to the positions where two hashes disagree: if two images are true duplicates, they differ in at most 10 of the 64 bits. Because the 6 chunks don't overlap and together cover all 64 bits, the total difference is just the sum of the per-chunk differences. So if every chunk differed by 2 or more bits, the total would be at least 2 × 6 = 12 — but a duplicate has at most 10 differences, and you can't have ≤10 and ≥12 at once. That contradiction forces at least one chunk down to at most floor(10/6) = 1 differing bit. In other words, for any real duplicate, there is guaranteed to be at least one chunk where the two hashes are within 1 bit of each other.

So we index every kept image by all 6 of its chunk values, and for each new pHash we split it into 6 chunks and probe all of them — each at its exact value plus every value within chunk_radius = 1 bit (the variant masks). We check all 6 because pigeonhole guarantees some chunk is close but not which one. Any kept image found in those buckets is a candidate, which we then confirm with an exact Hamming-distance check on the full hash (sharing one near-chunk is necessary but not sufficient, since the other chunks could still differ a lot). If none of the 6 chunks produces a match, the image is guaranteed not to be a duplicate — because a real duplicate would have been forced to leave at least one near-matching chunk behind.

The chunk lookups are dictionary (hash-map) lookups, which jump **directly** to the drawer for a given chunk value without scanning the other drawers. So for each new pHash we do a fixed number of cheap lookups (6 chunks × the variant masks, a constant that doesn't grow with the dataset), and those lookups hand us only the small set of kept items that share a near-chunk — the candidates. We run the expensive full-hash Hamming check **only** on those candidates, never on the rest. Every kept item filed under unrelated chunk values is never touched at all. Naive dedup is O(n) comparisons per new item (compare against all n kept items), so the whole run is O(n²); the index replaces that per-item scan with a handful of constant-time lookups plus a few exact checks on candidates, so in practice each item costs roughly O(1) instead of O(n) — turning an O(n²) algorithm into something close to O(n). The pigeonhole guarantee is what makes this safe: since a real duplicate is forced to share a near-chunk, the items we skip are provably not duplicates, so ignoring them costs us no correctness.

edwinpav (approved):

vinay553:
@edwinpav I think your explanation is pretty good. To be pedantic, it's not necessarily most and the runtime stuff is handwavy and really can only be validated empirically. Some other key points: 
The size of the chunks is key. I picked 10 as a reasonable upper limit for the threshold. So given 10 we needed to keep the chunks exactly this size so the floor(10/6) = 1, because if it was floor(10/N) > 1, the amount of adjustments scale exponentially. 
The other thing is the bit rotation to narrow candidates down further.

nucleus/local_deduplication.py
vinay553: I left this local-only for now. The existing async `Dataset.deduplicate()` path relies on backend/job validation today, and the existing integration tests expect `NucleusAPIError`/`JobError` for invalid thresholds. Adding shared client-side validation there would change that public error behavior. This local helper needs the upfront check because it does not make an API call and should fail before doing CPU-heavy local work.

nucleus/local_deduplication.py
vinay553: I did not add this because the supported local input shape only exposes a `DatasetItem`, and the SDK currently discards the backend item ID when it builds `DatasetItem` objects from export/API payloads. `reference_id` is the only reliable identifier available here; adding a best-effort `dataset_item_id` field would usually be absent for the exact generator rows this utility is intended to consume.

nucleus/local_deduplication.py
vinay553: Done, the empty-input return now happens immediately after normalization and before the sort.

edwinpav:
> @edwinpav I think your explanation is pretty good. To be pedantic, it's not necessarily most and the runtime stuff is handwavy and really can only be validated empirically. Some other key points: The size of the chunks is key. I picked 10 as a reasonable upper limit for the threshold. So given 10 we needed to keep the chunks exactly this size so the floor(10/6) = 1, because if it was floor(10/N) > 1, the amount of adjustments scale exponentially. The other thing is the bit rotation to narrow candidates down further.

My understanding was that the bit rotation is because since the floor(10/n) = 1, we need to rotate each candidate chunk by 1 bit in each position to check if its 1 bit difference but any stores ones in that chunk. Or is it also/just for something else?

edwinpav:
also @vinay553 if u [follow this](https://github.com/scaleapi/nucleus-python-client#for-developers) and run `poetry run pre-commit install` it'll run all lint checks for you locally on each commit so you can fix them before ci/cd fails on it

vinay553:
@edwinpav ah I wouldn't call that rotating, that's just applying a mask. rotation is when you rotate the bits like 1234 -> 2341. The reason is that any potential duplicate candidate would need to pass the chunk test both times before we do the full 64-bit check, reducing then number of candidates.

edwinpav:
> @edwinpav ah I wouldn't call that rotating, that's just applying a mask. rotation is when you rotate the bits like 1234 -> 2341. The reason is that any potential duplicate candidate would need to pass the chunk test both times before we do the full 64-bit check, reducing then number of candidates.

I see, why is the rotation safe? ie. we guaranteed that it wont be a duplicate if they dont match?

vinay553:
> > @edwinpav ah I wouldn't call that rotating, that's just applying a mask. rotation is when you rotate the bits like 1234 -> 2341. The reason is that any potential duplicate candidate would need to pass the chunk test both times before we do the full 64-bit check, reducing then number of candidates.
> 
> I see, why is the rotation safe? ie. we guaranteed that it wont be a duplicate if they dont match?

We can talk abt the algo in our 1:1

PR #464, opened 2026-06-08
[codex] Add native pHash deduplication acceleration

vinay553:
## Summary

Adds native C acceleration for the local `deduplicate_by_phash` utility and releases it as `0.18.5`.

The public SDK API is unchanged. When `nucleus._native_dedup` is available, `deduplicate_by_phash` sends the sorted pHash values to native code and reconstructs the existing `LocalDeduplicationResult` from the returned kept indexes. If the extension is unavailable, the existing pure-Python implementation is still used.

## Native Threshold Routing

Native code now handles every valid threshold:

- `0 <= threshold <= 11`: chunked Hamming index
- `12 <= threshold <= 63`: native linear scan using `uint64_t` XOR + popcount
- `threshold == 64`: keep-first fast path

The threshold-index cap moves from `10` to `11` because `floor(11 / 6) == 1`, so it has the same one-bit chunk-radius behavior as threshold `10`.

## Packaging / CI Notes

This adds a Poetry build script backed by `setuptools.Extension`. The extension is marked `optional=True`, so environments that cannot compile the C module can still build/install the SDK and use the Python fallback.

The CircleCI tag-release workflow now builds release artifacts in separate jobs and publishes the combined `dist/` directory:

- source distribution
- Linux `x86_64` wheels for CPython 3.10 through 3.14
- macOS `universal2` wheels for CPython 3.10 through 3.14
- Windows `amd64` wheels for CPython 3.10 through 3.14

Each wheel job uses `cibuildwheel` and runs a smoke test that imports `nucleus._native_dedup` and verifies an indexed threshold result. The final `pypi_publish` job attaches the wheel/sdist workspace and uses the existing PyPI credentials context before running `poetry publish`.

## Validation

- `NUCLEUS_PYTEST_API_KEY=dummy poetry run pytest tests/test_local_deduplication.py -q`
- `NUCLEUS_PYTEST_API_KEY=dummy poetry run pre-commit run --all-files`
- `poetry build`
- `poetry run python -m pip wheel --no-deps --no-cache-dir -w /tmp/nucleus-native-0185-sdist-wheel dist/scale_nucleus-0.18.5.tar.gz`
- Installed local wheel and smoke-tested thresholds `0`, `10`, `11`, `12`, and `64` through the public API, plus direct native threshold `12` output.
- Parsed `.circleci/config.yml` locally with PyYAML.

<h3>Greptile Summary</h3>

This PR adds a native C extension (`nucleus._native_dedup`) to accelerate pHash deduplication and releases it as version `0.18.6`. The public Python API is unchanged; when the compiled extension is present, all thresholds (0–64) are handled natively, with a pure-Python fallback otherwise. A new multi-job CircleCI pipeline builds platform wheels (Linux `x86_64`, macOS `universal2`, Windows `amd64`) for Python 3.10–3.14 using `cibuildwheel` and publishes them alongside the sdist.

- **C extension** (`nucleus/_native_dedup.c`): implements three paths — a two-partition chunked Hamming index for thresholds 0–11, a linear XOR+popcount scan for 12–63, and a keep-first fast path for threshold 64. Memory management and Python C API error handling are correct throughout.
- **Python integration** (`nucleus/local_deduplication.py`): wraps the import in `try/except ImportError`, passes sorted `phash_value` integers to the native function, and reconstructs `_DeduplicationRecord` results from the returned index list.
- **CI** (`.circleci/config.yml`): splits the old single `pypi_publish` job into `build_sdist`, three platform wheel builders, and a final `pypi_publish` that attaches the combined workspace before uploading.

Safe to merge. The C extension is mathematically correct, memory management is clean, and the Python fallback is properly guarded.

The two-partition chunked Hamming index holds its pigeonhole invariant for all handled thresholds (0–11), the linear scan and keep-first paths are straightforward, and every allocation/cleanup path in the C code is covered. The Python integration correctly wraps the import, passes sorted integer hashes, and reconstructs results from returned indexes. The CIBW_ environment-variable prefix is used correctly across all three platform wheel jobs.

.circleci/config.yml — the wheel jobs could be made to depend on build_sdist to avoid wasting CI minutes when the PyPI version check fails, but this does not affect correctness.
</details>

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| nucleus/_native_dedup.c | New C extension: three-path deduplication (chunked Hamming index, linear scan, keep-first). Memory allocation, error propagation, and cleanup are all correct; pigeonhole proof holds for chunk_radius guarantees. |
| nucleus/local_deduplication.py | Adds _deduplicate_with_native helper and bumps LOCAL_INDEX_MAX_THRESHOLD from 10 to 11 (consistent with C code). Native path is correctly short-circuited before the Python dispatch chain. |
| .circleci/config.yml | Adds four new jobs (build_sdist, three platform wheel builders, reorganised pypi_publish). CIBW_ prefix is used correctly in all three wheel jobs. Minor: wheel builds start in parallel with build_sdist, so they will run even if the sdist validation step fails. |
| build.py | Poetry build hook wires the C extension through setuptools with -O3 / -std=c11 on non-Windows and optional=True for graceful fallback. |
| tests/test_local_deduplication.py | New parametrised tests cover the native path via monkeypatching and, when the extension is built, directly validate index (10, 11) and linear-scan (12) thresholds. |

</details>

```mermaid
flowchart TD
    A[deduplicate_by_phash called] --> B[validate threshold]
    B --> C[normalize & sort records by phash_value, stable_id]
    C --> D{_NATIVE_DEDUP available?}
    D -- Yes --> E[_deduplicate_with_native]
    E --> F{threshold routing in C}
    F -- threshold == 64 --> G[deduplicate_keep_first]
    F -- 0 to 11 --> H[deduplicate_with_index two-partition chunked Hamming]
    F -- 12 to 63 --> I[deduplicate_with_linear_scan XOR + popcount]
    G & H & I --> J[return kept_indexes list]
    J --> K[reconstruct records from indexes]
    D -- No --> L{Python fallback dispatch}
    L -- threshold == 0 --> M[_deduplicate_exact]
    L -- threshold == 64 --> N[keep first record]
    L -- 1 to 11 --> O[_deduplicate_with_index Python HammingIndex]
    L -- 12 to 63 --> P[_deduplicate_with_linear_scan]
    K & M & N & O & P --> Q[build LocalDeduplicationResult]
```
</details>

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `.circleci/config.yml`, line 58-63 ([link](https://github.com/scaleapi/nucleus-python-client/blob/f054278a251c11194824098d9535a01bf10e4e64/.circleci/config.yml#L58-L63)) 

   <a href="#">P1</a> **Wrong `cibuildwheel` environment variable prefix**

   All `cibuildwheel` configuration variables use the `CIBW_` prefix (e.g., `CIBW_BUILD`, `CIBW_SKIP`, `CIBW_ARCHS_LINUX`, `CIBW_TEST_COMMAND`), not `CIB_`. With the wrong prefix the variables are silently ignored: `cibuildwheel` will build its default set of targets (all detected Python versions, including PyPy, and all default architectures), the musllinux skip won't apply, and the smoke test won't run. This affects all three wheel jobs (`build_linux_wheels`, `build_macos_wheels`, `build_windows_wheels`).

   Replace every `CIB_` prefix with `CIBW_` across all three wheel jobs.

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   This is a comment left during a code review.
   Path: .circleci/config.yml
   Line: 58-63

   Comment:
   **Wrong `cibuildwheel` environment variable prefix**

   All `cibuildwheel` configuration variables use the `CIBW_` prefix (e.g., `CIBW_BUILD`, `CIBW_SKIP`, `CIBW_ARCHS_LINUX`, `CIBW_TEST_COMMAND`), not `CIB_`. With the wrong prefix the variables are silently ignored: `cibuildwheel` will build its default set of targets (all detected Python versions, including PyPy, and all default architectures), the musllinux skip won't apply, and the smoke test won't run. This affects all three wheel jobs (`build_linux_wheels`, `build_macos_wheels`, `build_windows_wheels`).

   Replace every `CIB_` prefix with `CIBW_` across all three wheel jobs.

   How can I resolve this? If you propose a fix, please make it concise.
   `````
   </details>

   <a href="https://app.greptile.com/api/ide/cursor?prompt=This%20is%20a%20comment%20left%20during%20a%20code%20review.%0APath%3A%20.circleci%2Fconfig.yml%0ALine%3A%2058-63%0A%0AComment%3A%0A**Wrong%20%60cibuildwheel%60%20environment%20variable%20prefix**%0A%0AAll%20%60cibuildwheel%60%20configuration%20variables%20use%20the%20%60CIBW_%60%20prefix%20%28e.g.%2C%20%60CIBW_BUILD%60%2C%20%60CIBW_SKIP%60%2C%20%60CIBW_ARCHS_LINUX%60%2C%20%60CIBW_TEST_COMMAND%60%29%2C%20not%20%60CIB_%60.%20With%20the%20wrong%20prefix%20the%20variables%20are%20silently%20ignored%3A%20%60cibuildwheel%60%20will%20build%20its%20default%20set%20of%20targets%20%28all%20detected%20Python%20versions%2C%20including%20PyPy%2C%20and%20all%20default%20architectures%29%2C%20the%20musllinux%20skip%20won't%20apply%2C%20and%20the%20smoke%20test%20won't%20run.%20This%20affects%20all%20three%20wheel%20jobs%20%28%60build_linux_wheels%60%2C%20%60build_macos_wheels%60%2C%20%60build_windows_wheels%60%29.%0A%0AReplace%20every%20%60CIB_%60%20prefix%20with%20%60CIBW_%60%20across%20all%20three%20wheel%20jobs.%0A%0AHow%20can%20I%20resolve%20this%3F%20If%20you%20propose%20a%20fix%2C%20please%20make%20it%20concise.&pr=464&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInCursorDark.svg?v=3"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInCursor.svg?v=3">Fix in Cursor</picture></a> <a href="https://app.greptile.com/ide/claude-code?prompt=This%20is%20a%20comment%20left%20during%20a%20code%20review.%0APath%3A%20.circleci%2Fconfig.yml%0ALine%3A%2058-63%0A%0AComment%3A%0A**Wrong%20%60cibuildwheel%60%20environment%20variable%20prefix**%0A%0AAll%20%60cibuildwheel%60%20configuration%20variables%20use%20the%20%60CIBW_%60%20prefix%20%28e.g.%2C%20%60CIBW_BUILD%60%2C%20%60CIBW_SKIP%60%2C%20%60CIBW_ARCHS_LINUX%60%2C%20%60CIBW_TEST_COMMAND%60%29%2C%20not%20%60CIB_%60.%20With%20the%20wrong%20prefix%20the%20variables%20are%20silently%20ignored%3A%20%60cibuildwheel%60%20will%20build%20its%20default%20set%20of%20targets%20%28all%20detected%20Python%20versions%2C%20including%20PyPy%2C%20and%20all%20default%20architectures%29%2C%20the%20musllinux%20skip%20won't%20apply%2C%20and%20the%20smoke%20test%20won't%20run.%20This%20affects%20all%20three%20wheel%20jobs%20%28%60build_linux_wheels%60%2C%20%60build_macos_wheels%60%2C%20%60build_windows_wheels%60%29.%0A%0AReplace%20every%20%60CIB_%60%20prefix%20with%20%60CIBW_%60%20across%20all%20three%20wheel%20jobs.%0A%0AHow%20can%20I%20resolve%20this%3F%20If%20you%20propose%20a%20fix%2C%20please%20make%20it%20concise.&repo=scaleapi%2Fnucleus-python-client&pr=464&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInClaudeDark.svg?v=3"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInClaude.svg?v=3">Fix in Claude Code</picture></a> <a href="https://app.greptile.com/api/ide/codex?prompt=IMPORTANT%3A%20Work%20in%20the%20repository%20%22scaleapi%2Fnucleus-python-client%22%20on%20the%20existing%20branch%20%22codex%2Fnative-dedup-acceleration%22.%20Checkout%20that%20branch%20%E2%80%94%20do%20NOT%20create%20a%20new%20branch%20or%20open%20a%20new%20PR.%20Push%20your%20changes%20to%20%22codex%2Fnative-dedup-acceleration%22.%0A%0AThis%20is%20a%20comment%20left%20during%20a%20code%20review.%0APath%3A%20.circleci%2Fconfig.yml%0ALine%3A%2058-63%0A%0AComment%3A%0A**Wrong%20%60cibuildwheel%60%20environment%20variable%20prefix**%0A%0AAll%20%60cibuildwheel%60%20configuration%20variables%20use%20the%20%60CIBW_%60%20prefix%20%28e.g.%2C%20%60CIBW_BUILD%60%2C%20%60CIBW_SKIP%60%2C%20%60CIBW_ARCHS_LINUX%60%2C%20%60CIBW_TEST_COMMAND%60%29%2C%20not%20%60CIB_%60.%20With%20the%20wrong%20prefix%20the%20variables%20are%20silently%20ignored%3A%20%60cibuildwheel%60%20will%20build%20its%20default%20set%20of%20targets%20%28all%20detected%20Python%20versions%2C%20including%20PyPy%2C%20and%20all%20default%20architectures%29%2C%20the%20musllinux%20skip%20won't%20apply%2C%20and%20the%20smoke%20test%20won't%20run.%20This%20affects%20all%20three%20wheel%20jobs%20%28%60build_linux_wheels%60%2C%20%60build_macos_wheels%60%2C%20%60build_windows_wheels%60%29.%0A%0AReplace%20every%20%60CIB_%60%20prefix%20with%20%60CIBW_%60%20across%20all%20three%20wheel%20jobs.%0A%0AHow%20can%20I%20resolve%20this%3F%20If%20you%20propose%20a%20fix%2C%20please%20make%20it%20concise.&repo=scaleapi%2Fnucleus-python-client&pr=464&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInCodexDark.svg?v=3"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInCodex.svg?v=3">Fix in Codex</picture></a>
</details>

<a href="https://app.greptile.com/api/ide/cursor?prompt=Fix%20the%20following%201%20code%20review%20issue.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0A.circleci%2Fconfig.yml%3A295-322%0A**Wheel%20builds%20run%20even%20when%20sdist%20validation%20fails**%0A%0A%60build_linux_wheels%60%2C%20%60build_macos_wheels%60%2C%20and%20%60build_windows_wheels%60%20all%20require%20only%20%60build_test%60%2C%20not%20%60build_sdist%60.%20The%20%60build_sdist%60%20job%20performs%20the%20%22Validate%20SDK%20Version%20Increment%22%20check%20that%20exits%20early%20if%20the%20version%20already%20exists%20on%20PyPI%20%E2%80%94%20but%20since%20the%20three%20wheel%20jobs%20start%20in%20parallel%20with%20%60build_sdist%60%2C%20they%20will%20spin%20up%20and%20consume%20CI%20time%20even%20when%20that%20validation%20would%20have%20blocked%20the%20release.%20%60pypi_publish%60%20requires%20all%20four%2C%20so%20publishing%20is%20still%20blocked%2C%20but%20the%20wheel-build%20minutes%20are%20wasted.%0A%0AAdding%20%60build_sdist%60%20to%20each%20wheel%20job's%20%60requires%60%20list%20would%20short-circuit%20the%20pipeline%20on%20a%20validation%20failure%20before%20wheel%20builds%20start.%0A%0A&pr=464&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCursorDark.svg?v=3"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCursor.svg?v=3">Fix All in Cursor</picture></a> <a href="https://app.greptile.com/ide/claude-code?prompt=Fix%20the%20following%201%20code%20review%20issue.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0A.circleci%2Fconfig.yml%3A295-322%0A**Wheel%20builds%20run%20even%20when%20sdist%20validation%20fails**%0A%0A%60build_linux_wheels%60%2C%20%60build_macos_wheels%60%2C%20and%20%60build_windows_wheels%60%20all%20require%20only%20%60build_test%60%2C%20not%20%60build_sdist%60.%20The%20%60build_sdist%60%20job%20performs%20the%20%22Validate%20SDK%20Version%20Increment%22%20check%20that%20exits%20early%20if%20the%20version%20already%20exists%20on%20PyPI%20%E2%80%94%20but%20since%20the%20three%20wheel%20jobs%20start%20in%20parallel%20with%20%60build_sdist%60%2C%20they%20will%20spin%20up%20and%20consume%20CI%20time%20even%20when%20that%20validation%20would%20have%20blocked%20the%20release.%20%60pypi_publish%60%20requires%20all%20four%2C%20so%20publishing%20is%20still%20blocked%2C%20but%20the%20wheel-build%20minutes%20are%20wasted.%0A%0AAdding%20%60build_sdist%60%20to%20each%20wheel%20job's%20%60requires%60%20list%20would%20short-circuit%20the%20pipeline%20on%20a%20validation%20failure%20before%20wheel%20builds%20start.%0A%0A&repo=scaleapi%2Fnucleus-python-client&pr=464&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=3"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=3">Fix All in Claude Code</picture></a> <a href="https://app.greptile.com/api/ide/codex?prompt=IMPORTANT%3A%20Work%20in%20the%20repository%20%22scaleapi%2Fnucleus-python-client%22%20on%20the%20existing%20branch%20%22codex%2Fnative-dedup-acceleration%22.%20Checkout%20that%20branch%20%E2%80%94%20do%20NOT%20create%20a%20new%20branch%20or%20open%20a%20new%20PR.%20Push%20your%20changes%20to%20%22codex%2Fnative-dedup-acceleration%22.%0A%0AFix%20the%20following%201%20code%20review%20issue.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0A.circleci%2Fconfig.yml%3A295-322%0A**Wheel%20builds%20run%20even%20when%20sdist%20validation%20fails**%0A%0A%60build_linux_wheels%60%2C%20%60build_macos_wheels%60%2C%20and%20%60build_windows_wheels%60%20all%20require%20only%20%60build_test%60%2C%20not%20%60build_sdist%60.%20The%20%60build_sdist%60%20job%20performs%20the%20%22Validate%20SDK%20Version%20Increment%22%20check%20that%20exits%20early%20if%20the%20version%20already%20exists%20on%20PyPI%20%E2%80%94%20but%20since%20the%20three%20wheel%20jobs%20start%20in%20parallel%20with%20%60build_sdist%60%2C%20they%20will%20spin%20up%20and%20consume%20CI%20time%20even%20when%20that%20validation%20would%20have%20blocked%20the%20release.%20%60pypi_publish%60%20requires%20all%20four%2C%20so%20publishing%20is%20still%20blocked%2C%20but%20the%20wheel-build%20minutes%20are%20wasted.%0A%0AAdding%20%60build_sdist%60%20to%20each%20wheel%20job's%20%60requires%60%20list%20would%20short-circuit%20the%20pipeline%20on%20a%20validation%20failure%20before%20wheel%20builds%20start.%0A%0A&repo=scaleapi%2Fnucleus-python-client&pr=464&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodexDark.svg?v=3"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodex.svg?v=3">Fix All in Codex</picture></a>

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
Fix the following 1 code review issue. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 1
.circleci/config.yml:295-322
**Wheel builds run even when sdist validation fails**

`build_linux_wheels`, `build_macos_wheels`, and `build_windows_wheels` all require only `build_test`, not `build_sdist`. The `build_sdist` job performs the "Validate SDK Version Increment" check that exits early if the version already exists on PyPI — but since the three wheel jobs start in parallel with `build_sdist`, they will spin up and consume CI time even when that validation would have blocked the release. `pypi_publish` requires all four, so publishing is still blocked, but the wheel-build minutes are wasted.

Adding `build_sdist` to each wheel job's `requires` list would short-circuit the pipeline on a validation failure before wheel builds start.


`````

</details>

<sub>Reviews (4): Last reviewed commit: ["Address native dedup review nits"](https://github.com/scaleapi/nucleus-python-client/commit/fc377e0b4f0c20d4d583c371210f8c150f88a610) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36329623)</sub>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,7 @@ version: 2.1
 orbs:
   slack: circleci/slack@4.4.2
   python: circleci/python@2.1.0
+  win: circleci/windows@5.0
 
 jobs:
   build_test:
@@ -75,11 +76,16 @@ jobs:
           event: fail
           template: basic_fail_1
 
-  pypi_publish:
+  build_sdist:
     docker:
       - image: cimg/python:3.10
     steps:
       - checkout # checkout source code to working directory
+      - run:
+          name: Install Build Tools
+          command: |
+            pip install --upgrade pip
+            pip install poetry
       - run:
           name: Validate Tag Version # Check if the tag name matches the package version
           command: |
@@ -99,9 +105,121 @@ jobs:
                 exit 1;
             fi
       - run:
-          name: Build
+          name: Build sdist
           command: | # install env dependencies
-            poetry build
+            rm -rf dist
+            poetry build --format sdist
+      - persist_to_workspace:
+          root: .
+          paths:
+            - dist
+
+  build_linux_wheels:
+    docker:
+      - image: cimg/python:3.10
+    steps:
+      - checkout # checkout source code to working directory
+      - setup_remote_docker
+      - run:
+          name: Build Linux wheels
+          command: |
+            pip install --upgrade pip
+            pip install cibuildwheel
+
+            export CIBW_ARCHS_LINUX="x86_64"
+            export CIBW_BUILD="cp310-manylinux_x86_64 cp311-manylinux_x86_64 cp312-manylinux_x86_64 cp313-manylinux_x86_64 cp314-manylinux_x86_64"
+            export CIBW_SKIP="pp* *-musllinux_*"
+            export CIBW_TEST_COMMAND='python -c "import nucleus._native_dedup as native; assert native.deduplicate_phashes([0, 1023, 2047], 10) == [0, 2]"'
+            python -m cibuildwheel --platform linux --output-dir dist
+
+            ls -lh dist
+      - persist_to_workspace:
+          root: .
+          paths:
+            - dist
+
+  build_macos_wheels:
+    macos:
+      xcode: 16.4.0
+    resource_class: m4pro.medium
+    steps:
+      - checkout # checkout source code to working directory
+      - run:
+          name: Build macOS wheels
+          command: |
+            python3 -m pip install --upgrade pip
+            python3 -m pip install cibuildwheel
+            softwareupdate --install-rosetta --agree-to-license || true
+
+            export CIBW_ARCHS_MACOS="universal2"
+            export CIBW_BUILD="cp310-macosx_* cp311-macosx_* cp312-macosx_* cp313-macosx_* cp314-macosx_*"
+            export CIBW_SKIP="pp*"
+            export CIBW_TEST_COMMAND='python -c "import nucleus._native_dedup as native; assert native.deduplicate_phashes([0, 1023, 2047], 10) == [0, 2]"'
+            python3 -m cibuildwheel --platform macos --output-dir dist
+
+            ls -lh dist
+      - persist_to_workspace:
+          root: .
+          paths:
+            - dist
+
+  build_windows_wheels:
+    executor:
+      name: win/default
+      size: medium
+    steps:
+      - checkout # checkout source code to working directory
+      - run:
+          name: Build Windows wheels
+          command: |
+            python -m pip install --upgrade pip
+            python -m pip install cibuildwheel
+
+            $env:CIBW_BUILD = "cp310-win_amd64 cp311-win_amd64 cp312-win_amd64 cp313-win_amd64 cp314-win_amd64"
+            $env:CIBW_SKIP = "pp*"
+            $env:CIBW_TEST_COMMAND = 'python -c "import nucleus._native_dedup as native; assert native.deduplicate_phashes([0, 1023, 2047], 10) == [0, 2]"'
+            python -m cibuildwheel --platform windows --output-dir dist
+
+            Get-ChildItem dist
+      - persist_to_workspace:
+          root: .
+          paths:
+            - dist
+
+  pypi_publish:
+    docker:
+      - image: cimg/python:3.10
+    steps:
+      - checkout # checkout source code to working directory
+      - attach_workspace:
+          at: .
+      - run:
+          name: Install Publish Tools
+          command: |
+            pip install --upgrade pip
+            pip install poetry
+      - run:
+          name: Validate Tag Version # Check if the tag name matches the package version
+          command: |
+            PKG_VERSION=$(sed -n 's/^version = //p' pyproject.toml | sed -e 's/^"//' -e 's/"$//')
+            if [[ "$CIRCLE_TAG" != "v${PKG_VERSION}" ]]; then
+                echo "ERROR: Tag name ($CIRCLE_TAG) must match package version (v${PKG_VERSION})."
+                exit 1;
+            fi
+      - run:
+          name: Validate SDK Version Increment # Check if the version is already on PyPI
+          command: |
+            PKG_VERSION=$(sed -n 's/^version = //p' pyproject.toml | sed -e 's/^"//' -e 's/"$//')
+            if pip install "scale-nucleus>=${PKG_VERSION}" > /dev/null 2>&1;
+            then
+                echo "ERROR: You need to increment to a new version before publishing!"
+                echo "Version (${PKG_VERSION}) already exists on PyPI."
+                exit 1;
+            fi
+      - run:
+          name: List artifacts to publish
+          command: |
+            ls -lh dist
       - run:
           name: Publish to PyPI
           command: |
@@ -165,7 +283,8 @@ workflows:
           filters:
             tags:
               only: /^v\d+\.\d+\.\d+$/ # Runs only for tags with the format [v1.2.3]
-      - pypi_publish:
+      - build_sdist:
+          context: Nucleus
           requires:
             - build_test
           filters:
@@ -173,4 +292,42 @@ workflows:
               ignore: /.*/ # Runs for none of the branches
             tags:
               only: /^v\d+\.\d+\.\d+$/ # Runs only for tags with the format [v1.2.3]
-
+      - build_linux_wheels:
+          context: Nucleus
+          requires:
+            - build_test
+          filters:
+            branches:
+              ignore: /.*/ # Runs for none of the branches
+            tags:
+              only: /^v\d+\.\d+\.\d+$/ # Runs only for tags with the format [v1.2.3]
+      - build_macos_wheels:
+          context: Nucleus
+          requires:
+            - build_test
+          filters:
+            branches:
+              ignore: /.*/ # Runs for none of the branches
+            tags:
+              only: /^v\d+\.\d+\.\d+$/ # Runs only for tags with the format [v1.2.3]
+      - build_windows_wheels:
+          context: Nucleus
+          requires:
+            - build_test
+          filters:
+            branches:
+              ignore: /.*/ # Runs for none of the branches
+            tags:
+              only: /^v\d+\.\d+\.\d+$/ # Runs only for tags with the format [v1.2.3]
+      - pypi_publish:
+          context: Nucleus
+          requires:
+            - build_sdist
+            - build_linux_wheels
+            - build_macos_wheels
+            - build_windows_wheels
+          filters:
+            branches:
+              ignore: /.*/ # Runs for none of the branches
+            tags:
+              only: /^v\d+\.\d+\.\d+$/ # Runs only for tags with the format [v1.2.3]
--- a/build.py
+++ b/build.py
@@ -1,0 +1,24 @@
+from __future__ import annotations
+
+import sys
+
+from setuptools import Extension
+
+
+def build(setup_kwargs):
+    extra_compile_args = []
+    if sys.platform != "win32":
+        extra_compile_args.extend(["-std=c11", "-O3"])
+
+    setup_kwargs.update(
+        {
+            "ext_modules": [
+                Extension(
+                    "nucleus._native_dedup",
+                    ["nucleus/_native_dedup.c"],
+                    extra_compile_args=extra_compile_args,
+                    optional=True,
+                )
+            ],
+        }
+    )
--- a/nucleus/_native_dedup.c
+++ b/nucleus/_native_dedup.c
@@ -17,6 +17,12 @@
 
 static const int CHUNK_BITS[CHUNK_COUNT] = {11, 11, 11, 11, 10, 10};
 
+typedef enum {
+    CANDIDATE_MARK_UNMARKED = 0,   /* default; not involved in this query */
+    CANDIDATE_MARK_CANDIDATE = 1,  /* partition 0 hit; check in partition 1 */
+    CANDIDATE_MARK_REJECTED = 2,   /* partition 1 checked; not within threshold */
+} CandidateMark;
+
 typedef struct {
     Py_ssize_t *values;
     Py_ssize_t size;
@@ -128,7 +134,8 @@ static void clear_candidate_marks(HammingIndex *index) {
     Py_ssize_t i;
 
     for (i = 0; i < index->touched_count; i++) {
-        index->candidate_marks[index->touched_indexes[i]] = 0;
+        index->candidate_marks[index->touched_indexes[i]] =
+            CANDIDATE_MARK_UNMARKED;
     }
     index->touched_count = 0;
 }
@@ -206,8 +213,8 @@ static int mark_bucket_candidates(HammingIndex *index, Bucket *bucket) {
 
     for (i = 0; i < bucket->size; i++) {
         kept_index = bucket->values[i];
-        if (index->candidate_marks[kept_index] == 0) {
-            index->candidate_marks[kept_index] = 1;
+        if (index->candidate_marks[kept_index] == CANDIDATE_MARK_UNMARKED) {
+            index->candidate_marks[kept_index] = CANDIDATE_MARK_CANDIDATE;
             if (touched_append(index, kept_index) < 0) {
                 return -1;
             }
@@ -286,14 +293,15 @@ static int find_partition_one_duplicate(
             for (bucket_offset = 0; bucket_offset < bucket->size;
                  bucket_offset++) {
                 kept_index = bucket->values[bucket_offset];
-                if (index->candidate_marks[kept_index] != 1) {
+                if (index->candidate_marks[kept_index] !=
+                    CANDIDATE_MARK_CANDIDATE) {
                     continue;
                 }
                 if (popcount64(phash_value ^ index->hashes[kept_index]) <=
                     index->threshold) {
                     return 1;
                 }
-                index->candidate_marks[kept_index] = 2;
+                index->candidate_marks[kept_index] = CANDIDATE_MARK_REJECTED;
             }
         }
     }
@@ -337,7 +345,7 @@ static int hamming_index_add(HammingIndex *index, uint64_t phash_value) {
     }
 
     index->hashes[kept_index] = phash_value;
-    index->candidate_marks[kept_index] = 0;
+    index->candidate_marks[kept_index] = CANDIDATE_MARK_UNMARKED;
     for (partition_index = 0; partition_index < PARTITION_COUNT;
          partition_index++) {
         partition_chunks(phash_value, partition_index, chunks);
--- a/nucleus/_native_dedup.c
+++ b/nucleus/_native_dedup.c
@@ -1,0 +1,595 @@
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#if defined(_MSC_VER)
+#include <intrin.h>
+#endif
+
+#define DEDUP_THRESHOLD_MAX 64
+#define INDEX_MAX_THRESHOLD 11
+#define PARTITION_COUNT 2
+#define CHUNK_COUNT 6
+#define ROTATED_PARTITION_BITS 8
+
+static const int CHUNK_BITS[CHUNK_COUNT] = {11, 11, 11, 11, 10, 10};
+
+typedef struct {
+    Py_ssize_t *values;
+    Py_ssize_t size;
+    Py_ssize_t capacity;
+} Bucket;
+
+typedef struct {
+    Bucket *buckets;
+    Py_ssize_t bucket_count;
+} ChunkIndex;
+
+typedef struct {
+    int threshold;
+    int chunk_radius;
+    uint64_t *hashes;
+    uint8_t *candidate_marks;
+    Py_ssize_t kept_count;
+    Py_ssize_t hash_capacity;
+    Py_ssize_t *touched_indexes;
+    Py_ssize_t touched_count;
+    Py_ssize_t touched_capacity;
+    ChunkIndex indexes[PARTITION_COUNT][CHUNK_COUNT];
+} HammingIndex;
+
+static int popcount64(uint64_t value) {
+#if defined(_MSC_VER)
+    return (int)__popcnt64(value);
+#elif defined(__GNUC__) || defined(__clang__)
+    return __builtin_popcountll(value);
+#else
+    int count = 0;
+    while (value != 0) {
+        value &= value - 1;
+        count++;
+    }
+    return count;
+#endif
+}
+
+static uint64_t rotate_right_64(uint64_t value, int bits) {
+    return (value >> bits) | (value << (64 - bits));
+}
+
+static void partition_chunks(
+    uint64_t phash_value, int partition_index, uint64_t chunks[CHUNK_COUNT]
+) {
+    int chunk_index;
+    int shift = 64;
+
+    if (partition_index == 1) {
+        phash_value = rotate_right_64(phash_value, ROTATED_PARTITION_BITS);
+    }
+
+    for (chunk_index = 0; chunk_index < CHUNK_COUNT; chunk_index++) {
+        shift -= CHUNK_BITS[chunk_index];
+        chunks[chunk_index] =
+            (phash_value >> shift) &
+            ((((uint64_t)1) << CHUNK_BITS[chunk_index]) - 1);
+    }
+}
+
+static int append_py_ssize(
+    Py_ssize_t **values,
+    Py_ssize_t *size,
+    Py_ssize_t *capacity,
+    Py_ssize_t value
+) {
+    Py_ssize_t new_capacity;
+    Py_ssize_t *new_values;
+
+    if (*size == *capacity) {
+        new_capacity = *capacity == 0 ? 4 : *capacity * 2;
+        if (new_capacity < *capacity) {
+            PyErr_NoMemory();
+            return -1;
+        }
+        new_values = (Py_ssize_t *)realloc(
+            *values, (size_t)new_capacity * sizeof(Py_ssize_t)
+        );
+        if (new_values == NULL) {
+            PyErr_NoMemory();
+            return -1;
+        }
+        *values = new_values;
+        *capacity = new_capacity;
+    }
+
+    (*values)[*size] = value;
+    *size += 1;
+    return 0;
+}
+
+static int bucket_append(Bucket *bucket, Py_ssize_t kept_index) {
+    return append_py_ssize(
+        &bucket->values, &bucket->size, &bucket->capacity, kept_index
+    );
+}
+
+static int touched_append(HammingIndex *index, Py_ssize_t kept_index) {
+    return append_py_ssize(
+        &index->touched_indexes,
+        &index->touched_count,
+        &index->touched_capacity,
+        kept_index
+    );
+}
+
+static void clear_candidate_marks(HammingIndex *index) {
+    Py_ssize_t i;
+
+    for (i = 0; i < index->touched_count; i++) {
+        index->candidate_marks[index->touched_indexes[i]] = 0;
+    }
+    index->touched_count = 0;
+}
+
+static int hamming_index_init(
+    HammingIndex *index, Py_ssize_t hash_capacity, int threshold
+) {
+    int partition_index;
+    int chunk_index;
+    Py_ssize_t bucket_count;
+    Py_ssize_t allocation_capacity = hash_capacity > 0 ? hash_capacity : 1;
+
+    memset(index, 0, sizeof(*index));
+    index->threshold = threshold;
+    index->chunk_radius = threshold / CHUNK_COUNT;
+    index->hash_capacity = allocation_capacity;
+    index->hashes = (uint64_t *)calloc(
+        (size_t)allocation_capacity, sizeof(uint64_t)
+    );
+    index->candidate_marks = (uint8_t *)calloc(
+        (size_t)allocation_capacity, sizeof(uint8_t)
+    );
+    if (index->hashes == NULL || index->candidate_marks == NULL) {
+        PyErr_NoMemory();
+        return -1;
+    }
+
+    for (partition_index = 0; partition_index < PARTITION_COUNT;
+         partition_index++) {
+        for (chunk_index = 0; chunk_index < CHUNK_COUNT; chunk_index++) {
+            bucket_count = ((Py_ssize_t)1) << CHUNK_BITS[chunk_index];
+            index->indexes[partition_index][chunk_index].bucket_count =
+                bucket_count;
+            index->indexes[partition_index][chunk_index].buckets =
+                (Bucket *)calloc((size_t)bucket_count, sizeof(Bucket));
+            if (index->indexes[partition_index][chunk_index].buckets == NULL) {
+                PyErr_NoMemory();
+                return -1;
+            }
+        }
+    }
+
+    return 0;
+}
+
+static void hamming_index_free(HammingIndex *index) {
+    int partition_index;
+    int chunk_index;
+    Py_ssize_t bucket_index;
+    ChunkIndex *chunk_index_ptr;
+
+    for (partition_index = 0; partition_index < PARTITION_COUNT;
+         partition_index++) {
+        for (chunk_index = 0; chunk_index < CHUNK_COUNT; chunk_index++) {
+            chunk_index_ptr = &index->indexes[partition_index][chunk_index];
+            if (chunk_index_ptr->buckets == NULL) {
+                continue;
+            }
+            for (bucket_index = 0; bucket_index < chunk_index_ptr->bucket_count;
+                 bucket_index++) {
+                free(chunk_index_ptr->buckets[bucket_index].values);
+            }
+            free(chunk_index_ptr->buckets);
+        }
+    }
+
+    free(index->hashes);
+    free(index->candidate_marks);
+    free(index->touched_indexes);
+}
+
+static int mark_bucket_candidates(HammingIndex *index, Bucket *bucket) {
+    Py_ssize_t i;
+    Py_ssize_t kept_index;
+
+    for (i = 0; i < bucket->size; i++) {
+        kept_index = bucket->values[i];
+        if (index->candidate_marks[kept_index] == 0) {
+            index->candidate_marks[kept_index] = 1;
+            if (touched_append(index, kept_index) < 0) {
+                return -1;
+            }
+        }
+    }
+    return 0;
+}
+
+static int mark_partition_zero_candidates(
+    HammingIndex *index, uint64_t phash_value
+) {
+    uint64_t chunks[CHUNK_COUNT];
+    int chunk_index;
+    int bit_index;
+    uint64_t chunk_value;
+    uint64_t variant_value;
+    ChunkIndex *chunk_index_ptr;
+
+    partition_chunks(phash_value, 0, chunks);
+    for (chunk_index = 0; chunk_index < CHUNK_COUNT; chunk_index++) {
+        chunk_index_ptr = &index->indexes[0][chunk_index];
+        chunk_value = chunks[chunk_index];
+        if (mark_bucket_candidates(
+                index, &chunk_index_ptr->buckets[chunk_value]
+            ) < 0) {
+            return -1;
+        }
+
+        if (index->chunk_radius == 0) {
+            continue;
+        }
+
+        for (bit_index = 0; bit_index < CHUNK_BITS[chunk_index];
+             bit_index++) {
+            variant_value = chunk_value ^ (((uint64_t)1) << bit_index);
+            if (mark_bucket_candidates(
+                    index, &chunk_index_ptr->buckets[variant_value]
+                ) < 0) {
+                return -1;
+            }
+        }
+    }
+    return 0;
+}
+
+static int find_partition_one_duplicate(
+    HammingIndex *index, uint64_t phash_value
+) {
+    uint64_t chunks[CHUNK_COUNT];
+    int chunk_index;
+    int bit_index;
+    uint64_t chunk_value;
+    uint64_t variant_value;
+    ChunkIndex *chunk_index_ptr;
+    Bucket *bucket;
+    Py_ssize_t bucket_offset;
+    Py_ssize_t kept_index;
+
+    partition_chunks(phash_value, 1, chunks);
+    for (chunk_index = 0; chunk_index < CHUNK_COUNT; chunk_index++) {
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

File: nucleus/_native_dedup.c [hunk 1]
```diff
@@ -0,0 +1,595 @@
...
+#define PARTITION_COUNT 2
+#define CHUNK_COUNT 6
+#define ROTATED_PARTITION_BITS 8
+
+static const int CHUNK_BITS[CHUNK_COUNT] = {11, 11, 11, 11, 10, 10};
+
+typedef struct {
+    Py_ssize_t *values;
+    Py_ssize_t size;
+    Py_ssize_t capacity;
```
vinay553, line 29: These are bookkeeping variables needed to create any kind of dynamically sized list in python. Size is the current size of the list, capacity is the amount of memory allocated for this list, and values is actually just a pointer to the first value in the list. Subsequent elements in a list are stored as neighbors in the array so you can get the 5th element of the list by doing *(values + 5). C has a shortcut where you can do values[5] just like other languages, but putting this here to explain why the struct looks like this. Size and capacity need to be tracked in order to add new elements to the end and potentially re-allocate the memory for the array if the array doesn't fit in its memory slot anymore.  

In general in this PR there are a few times you'll see something like *x, x_size, x_capacity. These are the boiler plate variables needed to essentially keep a dynamically sized array whose name is x, you should interpret them as "we are storing a list as a variable named x" in cases where we don't know the size of the list ahead of time.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+    return (int)__popcnt64(value);
+#elif defined(__GNUC__) || defined(__clang__)
+    return __builtin_popcountll(value);
+#else
+    int count = 0;
+    while (value != 0) {
+        value &= value - 1;
+        count++;
+    }
+    return count;
```
vinay553, line 62: Neat and fast way of counting the number of 1s in a 64 bit number. Basically (number & (number - 1)) will drop the last set bit of that number. So for example: 
13: 1101
13 & 12 = 1101 & 1100 = 1100 (12) 
12 & 11 = 1100 & 1011 = 1000 (8)
8 & 7 = 1000 & 0111 = 0000 (0)

It's easy to see why the last bit gets dropped when you do (x & (x-1)) for odd numbers, and if you think about it a bit it's not hard to see why it also works for even numbers

It took 3 iterations to get to 0 so we know there's 3 bits that are 1 in 13. This is useful because finding the bitwise difference between 2 pHashes (a and b) is the same as finding the number of 1s in (a ^ b), a^b is very fast, and this technique is also fast.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+#endif
+}
+
+static uint64_t rotate_right_64(uint64_t value, int bits) {
+    return (value >> bits) | (value << (64 - bits));
```
vinay553, line 67: value >> bits — low bits fall off; upper bits shift down
value << (64 - bits) — those dropped bits move to the top
| — combine into one rotated value

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+typedef struct {
+    Py_ssize_t *values;
+    Py_ssize_t size;
+    Py_ssize_t capacity;
+} Bucket;
+
+typedef struct {
+    Bucket *buckets;
+    Py_ssize_t bucket_count;
+} ChunkIndex;
```
vinay553, line 35: List of a list of buckets. Bucket_count is either 2^10 or 2^11 depending on whether the chunk corresponds to a chunk size of 10 or 11 (see below comment for why)

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+
+typedef struct {
+    int threshold;
+    int chunk_radius;
+    uint64_t *hashes;
```
vinay553, line 40: Hashes are a list of kept hashes (see top comment to understand how *hashes is essentially List[int] hashes).

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+    uint8_t *candidate_marks;
+    Py_ssize_t kept_count;
+    Py_ssize_t hash_capacity;
+    Py_ssize_t *touched_indexes;
+    Py_ssize_t touched_count;
+    Py_ssize_t touched_capacity;
+    ChunkIndex indexes[PARTITION_COUNT][CHUNK_COUNT];
```
vinay553, line 47: For each partition, for each chunk, store an array of 1024/2048 (depending on whether the chunk is 10 or 11 bits long) where each index maps to a one combination of 10/11 bits (this array is called a ChunkIndex). Each element in this array is a Bucket which is a dynamically sized list. For example if we want to store that elements 235 and 876 in the kept array have the bitset 42 in chunk 2 in when not rotated, indexes[0][2][41] = [235, 876]. If we wanted to add element 900 to this list we could do indexes[0][2][41][size + 1] = 900. It might feel weird that we're storing a list of size 1024/2048 with mostly empty elements at the beginning but those sizes aren't that big at the end of the day and it's how python "sets" or other hashmaps work under the hood.   

This array is most of the "meat" of this struct and if you think about it is basically all you need to run this algorithm.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+    return count;
+#endif
+}
+
+static uint64_t rotate_right_64(uint64_t value, int bits) {
+    return (value >> bits) | (value << (64 - bits));
+}
+
+static void partition_chunks(
+    uint64_t phash_value, int partition_index, uint64_t chunks[CHUNK_COUNT]
```
vinay553, line 71: Loads 6 chunks of the correct sizes into chunks given the phash_value. C cannot easily return lists, so you typically pass in the list as a parameter (as is done here with chunks) and the function will load it up.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+        chunks[chunk_index] =
+            (phash_value >> shift) &
+            ((((uint64_t)1) << CHUNK_BITS[chunk_index]) - 1);
+    }
+}
+
+static int append_py_ssize(
+    Py_ssize_t **values,
+    Py_ssize_t *size,
+    Py_ssize_t *capacity,
```
vinay553, line 91: This is a generic helper for adding elements to the dynamically sized arrays in this (like lists in python or TS, you don't need to specify the size beforehand). It essentially adds elements to the end, but if we need to add more space to the struct we realloc. We pass in pointers to these 64 bit integers instead of the actual value because then we can adjust them directly in this function.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+                PyErr_NoMemory();
+                return -1;
+            }
+        }
+    }
+
+    return 0;
+}
+
+static void hamming_index_free(HammingIndex *index) {
```
vinay553, line 185: In C since you're running directly on the machine, you're responsible for clearing all the bits on the heap once you're program is done (stack bits are cleared automatically). Otherwise some other random program might accidentally read it which is a huge security issue and may mess with the other program. These are called memory leaks.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+static void clear_candidate_marks(HammingIndex *index) {
+    Py_ssize_t i;
+
+    for (i = 0; i < index->touched_count; i++) {
+        index->candidate_marks[index->touched_indexes[i]] = 0;
+    }
+    index->touched_count = 0;
+}
+
+static int hamming_index_init(
```
vinay553, line 144: Allocates all the memory needed for the hamming index which is the single datastruct that manages this algorithm.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+#endif
+
+#define DEDUP_THRESHOLD_MAX 64
+#define INDEX_MAX_THRESHOLD 11
+#define PARTITION_COUNT 2
+#define CHUNK_COUNT 6
+#define ROTATED_PARTITION_BITS 8
+
+static const int CHUNK_BITS[CHUNK_COUNT] = {11, 11, 11, 11, 10, 10};
+
```
vinay553, line 19: Values used in the candidates list of the HammingIndex. These are stored on each candidate.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+typedef struct {
+    Bucket *buckets;
+    Py_ssize_t bucket_count;
+} ChunkIndex;
+
+typedef struct {
+    int threshold;
+    int chunk_radius;
+    uint64_t *hashes;
+    uint8_t *candidate_marks;
```
vinay553, line 41: Stores one of CandidateMark for each kept value (hashes). See comment on that enum to understand what each mean.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+    Py_ssize_t kept_count;
+    Py_ssize_t hash_capacity;
+    Py_ssize_t *touched_indexes;
```
vinay553, line 44: List of all the indices in hashes whose candidates_marks were set to something non-zero while processing this iteration. Since candidate_marks is reset to 0 for every new element we look at, this speeds up the resetting it to 0 since we just have to go through whatever indices are in here (hashes can be super long).

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+    Py_ssize_t kept_index;
+
+    for (i = 0; i < bucket->size; i++) {
+        kept_index = bucket->values[i];
+        if (index->candidate_marks[kept_index] == 0) {
+            index->candidate_marks[kept_index] = 1;
+            if (touched_append(index, kept_index) < 0) {
+                return -1;
+            }
+        }
```
vinay553, line 222: This will happen if we run out of memory, very rare error. -1 is a classic "error" return type in C.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+    }
+    return 0;
+}
+
+static int mark_partition_zero_candidates(
```
vinay553, line 227: My hope is that you can understand what this function is doing(at least mechanically, it's ok if you don't understand the algo) from the above comments.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+            return -1;
+        }
+
+        if (index->chunk_radius == 0) {
+            continue;
+        }
+
+        for (bit_index = 0; bit_index < CHUNK_BITS[chunk_index];
+             bit_index++) {
+            variant_value = chunk_value ^ (((uint64_t)1) << bit_index);
```
vinay553, line 253: Bitwise operator way of perturbing each bit in case the radius > 0 (radius can only 0 or 1)

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    result = build_kept_index_list(kept_indexes, kept_count);
+
+cleanup:
+    hamming_index_free(&index);
+    free(kept_indexes);
+    return result;
+}
+
```
vinay553, line 448: All the code above is to support this indexing algo which is fast for low thresholds. Everything below is to handle other cases.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+static PyObject *deduplicate_with_linear_scan(
+    const uint64_t *phashes, Py_ssize_t input_count, int threshold
+) {
+    Py_ssize_t *kept_indexes = NULL;
+    Py_ssize_t i;
```
vinay553, line 453: This is the "naive" way of doing deduplication where you just compare all new elements to everything kept. Lmk if something doesn't makes sense

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+
+    result = build_kept_index_list(kept_indexes, kept_count);
+    free(kept_indexes);
+    return result;
+}
+
+static PyObject *deduplicate_keep_first(Py_ssize_t input_count) {
+    PyObject *result;
+    PyObject *index_obj;
+
```
vinay553, line 498: This is a "special case" when the threshold is 64 where we just return the first element.

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,595 @@
...
+    if (PyList_Append(result, index_obj) < 0) {
+        Py_DECREF(index_obj);
+        Py_DECREF(result);
+        return NULL;
+    }
+    Py_DECREF(index_obj);
+    return result;
+}
+
+static PyObject *native_deduplicate_phashes(
```
vinay553, line 518: entrypoint

vinay553 (commented):
The naive implementation (that we use for large thresholds) is hopefully straightforward. The index implementation is a little complicated from an algo perspective, so I would try to ignore the algo itself (assume the algorithm works) and focus on making sure this is a correct implementation of the algorithm. At a basic mechanical level we are: 
1. Storing an index of every chunk of 11 bits of each pHash. 
2. Storing an index of every chunk of 11 bits of each pHash shifted by 8 bits. 
3. For each new pHash, collect a list of all pHashes whose bits differ by the new one by less than 2 for ANY chunk
4. Among the list we got in the previous step narrow it down to those whose bits differ by the new one by less than 2 for ANY chunk in the rotated forms. 
5. Doing a full 64 bit check among the remaining candidates. 

These steps above aren't that complicated, but because you're not familiar with C there is a lot of boilerplate that may look confusing. I have added inline comments to explain this boiler plate but definitely lmk if something doesn't make sense.

vinay553:
@greptile-apps review this pr

tests/test_local_deduplication.py
```diff
@@ -223,3 +229,54 @@ def test_deduplicate_by_phash_preserves_missing_reference_ids():
...
+)
+def test_native_deduplication_matches_python_index():
+    phashes = [
+        int("0" * 64, 2),
+        int(_flip_bits("0" * 64, range(10)), 2),
+        int(_flip_bits("0" * 64, range(11)), 2),
+        int("1" * 64, 2),
+    ]
+
+    assert _native_dedup.deduplicate_phashes(phashes, 10) == [0, 2, 3]
```
vinay553: Explicitly testing the _native_dedup.

nucleus/_native_dedup.c
edwinpav: This makes sense. My old C classes starting to come back to me lol. One small thing - you said "needed to create any kind of dynamically sized list in python" - do you mean "in c"?

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,603 @@
...
+#define CHUNK_COUNT 6
+#define ROTATED_PARTITION_BITS 8
+
+static const int CHUNK_BITS[CHUNK_COUNT] = {11, 11, 11, 11, 10, 10};
+
+typedef enum {
+    CANDIDATE_MARK_UNMARKED = 0,   /* default; not involved in this query */
+    CANDIDATE_MARK_CANDIDATE = 1,  /* partition 0 hit; check in partition 1 */
+    CANDIDATE_MARK_REJECTED = 2,   /* partition 1 checked; not within threshold */
+} CandidateMark;
```
edwinpav, line 24: nit: why aren't the enum values just `UNMARKED`, `CANDIDATE`, `REJECTED` if the name of the enum is already `CandidateMark`? Is the current form standard C syntax?

nucleus/_native_dedup.c
edwinpav: woah

nucleus/_native_dedup.c
edwinpav: So is my understanding correct? "indexes[0][2] slot 41 holding [235, 876] means "kept hashes #235 and #876 both have value 41 in chunk 2 of the un-rotated hash.""

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,603 @@
...
+    Py_ssize_t touched_capacity;
+    ChunkIndex indexes[PARTITION_COUNT][CHUNK_COUNT];
+} HammingIndex;
+
+static int popcount64(uint64_t value) {
+#if defined(_MSC_VER)
+    return (int)__popcnt64(value);
+#elif defined(__GNUC__) || defined(__clang__)
+    return __builtin_popcountll(value);
+#else
```
edwinpav, line 55: nit: maybe add that this is Kernighan's algorithm as a comment: `// Kernighan's algorithm`

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,603 @@
...
+
+    if (partition_index == 1) {
+        phash_value = rotate_right_64(phash_value, ROTATED_PARTITION_BITS);
+    }
+
+    for (chunk_index = 0; chunk_index < CHUNK_COUNT; chunk_index++) {
+        shift -= CHUNK_BITS[chunk_index];
+        chunks[chunk_index] =
+            (phash_value >> shift) &
+            ((((uint64_t)1) << CHUNK_BITS[chunk_index]) - 1);
```
edwinpav, line 84: so is this needed because we want just a chunk size of bits and remove all other bits that appear before this chunk? Isn't the chunk size of the `shifted hash AND mask` always equal to `shifted hash`?

build.py
```diff
@@ -0,0 +1,24 @@
...
+        extra_compile_args.extend(["-std=c11", "-O3"])
+
+    setup_kwargs.update(
+        {
+            "ext_modules": [
+                Extension(
+                    "nucleus._native_dedup",
+                    ["nucleus/_native_dedup.c"],
+                    extra_compile_args=extra_compile_args,
+                    optional=True,
```
edwinpav, line 20: It makes sense that this is optional and this probably won't happen, but curious once this is deployed, is it possible for you/the user who runs this dedup function to see if the C code is actually running or if its falling back to the python-based version. ie. could we detect if some weird/strange error occurred that causes the C code to never/barely run on user's who should be able to support C-mode?

.circleci/config.yml
```diff
@@ -75,11 +76,16 @@ jobs:
           event: fail
           template: basic_fail_1
 
-  pypi_publish:
+  build_sdist:
     docker:
       - image: cimg/python:3.10
```
edwinpav, line 81: nit: should this be the same as line 14? `- image: python:3.10-bullseye`

nucleus/local_deduplication.py
```diff
@@ -203,7 +210,10 @@ def deduplicate_by_phash(
 
     records.sort(key=lambda record: (record.phash_value, record.stable_id))
 
-    if threshold == 0:
+    native_unique_records = _deduplicate_with_native(records, threshold)
+    if native_unique_records is not None:
+        unique_records = native_unique_records
+    elif threshold == 0:
```
edwinpav, line 216: So if the native c dedup throws at runtime, we want it to error out and not fallback to python, right?

nucleus/_native_dedup.pyi
```diff
@@ -0,0 +1,5 @@
+from collections.abc import Iterable
+
+def deduplicate_phashes(
+    phashes: Iterable[int], threshold: int
```
edwinpav, line 4: seems like the c code uses `PySequence_Fast`, would `Sequence[int]` be better for type of `phashes`?

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,603 @@
...
+    const uint64_t *phashes, Py_ssize_t input_count, int threshold
+) {
+    HammingIndex index;
+    Py_ssize_t *kept_indexes = NULL;
+    Py_ssize_t i;
+    Py_ssize_t kept_count = 0;
+    int is_duplicate;
+    PyObject *result = NULL;
+
+    memset(&index, 0, sizeof(index));
```
edwinpav: Doesn't this happen inside of `hamming_index_init` already? This might be duplicated logic?

nucleus/_native_dedup.c
```diff
@@ -0,0 +1,603 @@
...
+    PyObject *phashes_obj;
+    PyObject *phashes_sequence;
+    PyObject **phash_items;
+    PyObject *result = NULL;
+    uint64_t *phashes = NULL;
+    Py_ssize_t input_count;
+    Py_ssize_t i;
+    int threshold;
+
+    (void)self;
```
edwinpav: instead of this, can you have the arg be `PyObject *Py_UNUSED(self)`?

edwinpav (approved):
Overall looks great, just some curiosity/nit comments. Also the greptile comments in the pr summary seem valid - about adding tests for the new index threshold of 11.

nucleus/local_deduplication.py
vinay553: Yeah, c only really throws for OOM which would be an issue in python as well

build.py
vinay553: Oh interesting, I don't know I'll look into it. I guess the in general clients don't have visibility into SDKs although nucleus-python-client is open source.

nucleus/_native_dedup.c
vinay553: yes

nucleus/_native_dedup.c
vinay553: the mask (line 84) removes the top bits, shift (line 83) only removes the bottom

nucleus/_native_dedup.pyi
vinay553: PySequence_Fast kind of a misnomer, it has nothing to do with Sequences. Since Sequences are a more constrained type than Iterable, I'm gonna leave it as is. you could pass any kind of iterable here and we'd be able to convert it to a PySequence_Fast object which is more permissive than a Sequence.

nucleus/_native_dedup.c
vinay553: yup sorry

nucleus/_native_dedup.c
vinay553: Hm it's sort of just C convention, but these are really just numbers, you can use them directly like CANDIDATE_MARK_UNMARKED, not CandidateMark.CANDIDATE_MARK_UNMARKED.

.circleci/config.yml
vinay553: nah it doesn't matter, the previous tests used the debian bullseye image so I just stuck to that convention

nucleus/_native_dedup.c
vinay553: good point

PR #469, opened 2026-07-27
[DE-8270] Model weights upload & download (SDK side)

luke-e-schaefer:
## Summary

SDK side of **model weights upload / download**, mirroring the REST surface merged in [scaleapi#149063](https://github.com/scaleapi/scaleapi/pull/149063).

The two primary methods are the ones the server PR's API-docs pages (`ApiDocsPage/models-python/{upload,download}-model-weights.md`) already document, so the published docs and the SDK agree:

```python
import nucleus

client = nucleus.NucleusClient(YOUR_SCALE_API_KEY)
model = client.get_model(reference_id="My-CNN")

client.upload_model_weights(model, "/path/to/weights.bin")
client.download_model_weights(model, "/path/to/save/weights.bin")
```

### Added

- **`NucleusClient.upload_model_weights(model, path, *, content_type=None, original_filename=None, checksum_sha256=None, on_progress=None)`** — presign → PUT direct to storage → finalize. Returns `ModelWeights`.
- **`NucleusClient.download_model_weights(model, path, *, on_progress=None)`** — resolves the signed URL and streams to disk (creating parent dirs). Returns the path written.
- **`get_model_weights(model)`** / **`delete_model_weights(model)`** for the remaining two routes.
- **`Model.upload_weights()` / `.download_weights()` / `.weights()` / `.delete_weights()`** — thin delegation to the client, matching how `Benchmark` wraps its client methods.
- **`ModelWeights`** metadata type: `present`, `status`, `size_bytes`, `original_filename`, `content_type`, `download_url`.

All model arguments accept either a `Model` or a bare model id (`prj_*`).

### Notes for review

- **Bytes never transit the Nucleus API.** Transfers go straight to storage over presigned URLs, so multi-GB artifacts aren't subject to API request-size limits.
- **Part PUTs are sent with no headers.** They're signed without the Content-Type condition, so forwarding `requiredHeaders` (which the *single* PUT does need) makes S3 reject the signature. There's a test pinning this in both directions — it's the easiest thing to get wrong here, and the frontend hook in 149063 has the same split.
- **Download uses `?json=1`** to fetch the signed URL rather than following the 302, so the API's auth headers are never sent to storage.
- **Multipart above 5 GB**, 4 parts in flight (a single S3 PUT is connection-throughput-bound). Missing part ETags fail on the first part rather than after transferring everything and dying at finalize.
- **The 10 GB server cap is checked client-side** before presign, so an oversized file fails without a network round-trip.
- The weights routes serialize **camelCase in both directions**, unlike most endpoints this SDK talks to, so the new payload keys are grouped and labelled as such in `constants.py`.

### Tests / Version

- `tests/test_model_weights.py` — 24 mock-based unit tests (no live API, no real S3): DTO parsing, payload builders, single vs. multipart transfer, header split, ETag/failure handling, progress callbacks, download streaming, all four client methods, and the `Model` wrappers.
- Verified locally the way CI does: `pylint nucleus` 10.00/10, `mypy --ignore-missing-imports nucleus` clean, ruff clean, black + isort clean, 61 mock-based tests passing across the eval/benchmark/leaderboard/weights suites.
- `pyproject.toml` → **0.19.1** + CHANGELOG entry (patch bump: additive new methods, per `CLAUDE.md`).

resolves https://linear.app/scale-epd/issue/DE-8270

<h3>Greptile Summary</h3>

This PR adds model weights upload and download to the Nucleus Python SDK, routing transfers directly to storage via presigned URLs so that multi-GB artifacts never pass through the Nucleus API. The implementation handles both single-PUT (< 5 GB) and S3 multipart (≥ 5 GB) paths with exponential-backoff retries, a threading-lock-protected progress counter for concurrent part uploads, atomic file replacement on download, and a `tqdm` progress bar that callers can suppress.

- **`NucleusClient`** gains `upload_model_weights`, `download_model_weights`, `get_model_weights`, and `delete_model_weights`; `Model` adds thin delegation wrappers (`upload_weights`, `download_weights`, `weights`, `delete_weights`) matching the existing `Benchmark` pattern.
- **`ModelWeights`** is the new metadata dataclass; `nucleus/constants.py` adds 14 camelCase wire-format keys (explicitly labelled as such since these routes do not snake-case their DTOs).
- 24 mock-based unit tests cover the full transfer surface including the per-part header split, ETag failure propagation, concurrency non-overlap, atomic download replacement, and all retry scenarios.

- Safe to merge. The transfer logic, retry handling, thread safety, and atomic file replacement are all correctly implemented and well-tested.
- The change is additive — four new client methods and one new module, with no modifications to existing code paths. The two minor findings (tqdm bar can show negative increments on a retried single-PUT, and multipart workers run to completion after a part error) are cosmetic/efficiency issues that don't affect data integrity or correctness. The core concerns from previous review rounds (race condition on the transferred counter, single-PUT progress reporting) have been addressed.
- nucleus/model_weights.py — specifically the _ProgressReader reset path and ThreadPoolExecutor shutdown behaviour on multipart failure, both noted in review comments.
</details>

| Filename | Overview |
|----------|----------|
| nucleus/model_weights.py | New core module implementing presign → PUT/multipart → finalize upload and streaming download with retry, progress reporting, and atomic file replacement. Logic is sound; two minor issues: progress bar can show negative deltas on single-PUT retry (bar.n not reset alongside _ProgressReader._transferred), and all multipart tasks run to completion even after a part fails due to ThreadPoolExecutor pool.map semantics. |
| nucleus/__init__.py | Adds upload_model_weights, download_model_weights, get_model_weights, and delete_model_weights to NucleusClient. Imports private helpers (_stream_weights_to_file etc.) from model_weights.py so they can be used inside the class body and patched in tests via nucleus.*. Implementation is clean and consistent with existing client patterns. |
| nucleus/model.py | Adds four thin delegation methods (upload_weights, download_weights, weights, delete_weights) to Model, exactly mirroring how Benchmark wraps client methods. No logic here. |
| nucleus/constants.py | Adds 14 camelCase constants for the model-weights wire format. The comment explaining why these are camelCase (unlike the rest of the SDK) is helpful. Constants are accurate representations of their values. |
| tests/test_model_weights.py | 24 well-structured unit tests covering DTO parsing, payload builders, single/multipart transfers, header split, ETag failure, progress callbacks (including the concurrency non-overlap test), retry/backoff, download streaming, atomic file replacement, and all four client methods plus Model wrappers. No live API or S3 calls. |

</details>

<details><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant User
    participant NucleusClient
    participant NucleusAPI
    participant Storage

    Note over User,Storage: Upload flow
    User->>NucleusClient: upload_model_weights(model, path)
    NucleusClient->>NucleusClient: os.path.getsize(path) → check ≤ 10 GB
    NucleusClient->>NucleusAPI: "POST model/{id}/weights/presign"
    NucleusAPI-->>NucleusClient: "{uploadId, uploadUrl|parts, requiredHeaders}"

    alt "Single PUT (< 5 GB)"
        NucleusClient->>Storage: PUT uploadUrl (with requiredHeaders, _ProgressReader)
        Storage-->>NucleusClient: ETag
    else Multipart (≥ 5 GB)
        par 4 concurrent workers
            NucleusClient->>Storage: PUT parts[n].url (no headers)
            Storage-->>NucleusClient: ETag[n]
        end
    end

    NucleusClient->>NucleusAPI: "POST model/{id}/weights/finalize {uploadId, parts}"
    NucleusAPI-->>NucleusClient: ModelWeights JSON
    NucleusClient-->>User: ModelWeights

    Note over User,Storage: Download flow
    User->>NucleusClient: download_model_weights(model, path)
    NucleusClient->>NucleusAPI: "GET model/{id}/weights/download?json=1"
    NucleusAPI-->>NucleusClient: "{url: signed-GET-URL}"
    NucleusClient->>Storage: GET signed-URL (streaming → temp file)
    Storage-->>NucleusClient: body chunks
    NucleusClient->>NucleusClient: os.replace(partial_path, path) [atomic]
    NucleusClient-->>User: path written
```
</details>

<sub>Reviews (11): Last reviewed commit: ["refactor(weights): drop the unused \`tota..."](https://github.com/scaleapi/nucleus-python-client/commit/f25061ac1bd659db0d00d4a9ee48a1b4eb6826de) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47721708)</sub>


nucleus/model_weights.py
```diff
@@ -0,0 +1,270 @@
...
+        if on_progress is not None:
+            on_progress(min(transferred, total_bytes), total_bytes)
+        return {PART_NUMBER_KEY: part_number, ETAG_KEY: etag}
+
+    with ThreadPoolExecutor(
+        max_workers=min(CONCURRENT_PART_UPLOADS, len(parts))
+    ) as pool:
+        finalized = list(pool.map(upload_part, parts))
+
+    return sorted(finalized, key=lambda p: p[PART_NUMBER_KEY])
```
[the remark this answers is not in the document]
luke-e-schaefer: Fixed in 2a8cad6. `transferred` and the value passed to `on_progress` are now both taken under a `threading.Lock`, so no part is dropped from the count. The callback itself still fires outside the lock, so delivery order across concurrent parts can interleave — the counter is what was broken, not the ordering.

nucleus/model_weights.py
```diff
@@ -0,0 +1,270 @@
...
+    path: str,
+    upload_url: str,
+    headers: Dict[str, str],
+    total_bytes: int,
+    on_progress: Optional[ProgressCallback],
+) -> None:
+    with open(path, "rb") as handle:
+        _put_bytes(upload_url, handle, headers)
+    if on_progress is not None:
+        on_progress(total_bytes, total_bytes)
```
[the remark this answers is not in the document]
luke-e-schaefer: Fixed in 2a8cad6 with the read-wrapper rather than the doc note. When `on_progress` is supplied, `_upload_single` wraps the handle in `_ProgressReader`, which counts bytes as requests reads them. It delegates everything but `read()`, so requests still sizes the body via `fileno()`/`tell()` and sends Content-Length instead of falling back to chunked — a chunked PUT would break the presigned signature. No callback means no wrapper, so the default path is byte-for-byte what it was. Covered by `test_transfer_single_put_reports_progress_while_the_body_is_read`.

edwinpav:
👀

nucleus/__init__.py
```diff
@@ -1685,6 +1698,156 @@ def delete_model(self, model_id: str) -> dict:
...
+                f"{MODEL_WEIGHTS_MAX_BYTES // 1024 ** 3} GB model weights limit"
+            )
+
+        presign = self.make_request(
+            _presign_payload(
+                total_bytes,
+                content_type,
+                original_filename
+                if original_filename is not None
+                else os.path.basename(path),
```
edwinpav: nit: assign to a var before this (maybe btwn line 1738 and 1739)

nucleus/__init__.py
```diff
@@ -1685,6 +1698,156 @@ def delete_model(self, model_id: str) -> dict:
...
+                checksum_sha256,
+            ),
+            f"model/{model_id}/weights/presign",
+        )
+        parts = _transfer_weights_to_storage(
+            path, presign, total_bytes, on_progress
+        )
```
edwinpav: should this be wrapped in a try/catch?

nucleus/__init__.py
```diff
@@ -1685,6 +1698,156 @@ def delete_model(self, model_id: str) -> dict:
...
+        finalized = self.make_request(
+            _finalize_payload(presign[UPLOAD_ID_KEY], parts),
```
edwinpav: presign[UPLOAD_ID_KEY] is unguarded. A presign response missing uploadId costs a complete 10 GB upload and then fails with a bare KeyError. Validate right after the presign call at 1746.

nucleus/__init__.py
```diff
@@ -1685,6 +1698,156 @@ def delete_model(self, model_id: str) -> dict:
         )
         return response
 
+    def upload_model_weights(
+        self,
+        model: Union[Model, str],
+        path: str,
```
edwinpav, line 1704: should path be ran through `os.path.expanduser` in both `upload_model_weights` and `download_model_weights`?

nucleus/__init__.py
```diff
@@ -1685,6 +1698,156 @@ def delete_model(self, model_id: str) -> dict:
...
+        # Ask for the URL as JSON rather than following the redirect, so the
+        # API credentials aren't replayed to the download host.
+        signed = self.make_request(
+            {},
+            f"model/{model_id}/weights/download?json=1",
+            requests_command=requests.get,
+        )
+        url = signed.get(URL_KEY)
+        if not url:
+            raise ValueError(
```
edwinpav: nit: does it make more sense for this to be `NotFoundError`? Up to you

nucleus/model_weights.py
```diff
@@ -0,0 +1,352 @@
...
+    """
+
+    model_project_id: Optional[str] = None
+    present: bool = False
+    status: Optional[str] = None
+    size_bytes: Optional[int] = None
+    original_filename: Optional[str] = None
+    content_type: Optional[str] = None
+    download_url: Optional[str] = None
+    _client: Optional["NucleusClient"] = field(repr=False, default=None)
```
edwinpav: is this ever used anywhere or is it not needed?

nucleus/model_weights.py
```diff
@@ -0,0 +1,352 @@
...
+    )
+    if not response.ok:
+        raise RuntimeError(
+            f"Upload to storage failed with status {response.status_code}: "
+            f"{response.text[:200]}"
+        )
+    return _strip_quotes(response.headers.get("ETag", ""))
+
+
+class _ProgressReader:
```
edwinpav, line 200: would it be more simple to wrap uploads within `tqdm` package which can natively show progress visually?

nucleus/model_weights.py
```diff
@@ -0,0 +1,352 @@
...
+    return etag.strip('"')
+
+
+def _read_part(path: str, offset: int, size: int) -> bytes:
+    with open(path, "rb") as handle:
+        handle.seek(offset)
+        return handle.read(size)
+
+
+def _put_bytes(
```
edwinpav, line 156: should there be some notion of retries for putting to s3?

nucleus/model_weights.py
```diff
@@ -0,0 +1,352 @@
...
+    # as the counter, not just the arithmetic: releasing first lets two threads
+    # compute 100 and 200 and then call in either order, so a caller's progress
+    # bar can jump backwards.
+    progress_lock = threading.Lock()
+
+    def upload_part(part: dict) -> Dict[str, Any]:
+        nonlocal transferred
+        part_number = int(part[PART_NUMBER_KEY])
+        offset = (part_number - 1) * part_size_bytes
+        chunk = _read_part(path, offset, part_size_bytes)
```
edwinpav, line 293: should there be validation that the chunk is correct length, check bad offsets, etc (or maybe do that within `_read_part`)

nucleus/model_weights.py
```diff
@@ -0,0 +1,352 @@
...
+        with progress_lock:
+            transferred += len(chunk)
+            if on_progress is not None:
+                on_progress(min(transferred, total_bytes), total_bytes)
+        return {PART_NUMBER_KEY: part_number, ETAG_KEY: etag}
+
+    with ThreadPoolExecutor(
+        max_workers=_part_upload_workers(len(parts), part_size_bytes)
+    ) as pool:
+        finalized = list(pool.map(upload_part, parts))
```
edwinpav, line 311: I think pool.map preserved input order so sorted() on line 239 might be redundant. worth a double check

nucleus/model_weights.py
```diff
@@ -0,0 +1,352 @@
...
+                    chunk_size=DOWNLOAD_CHUNK_BYTES
+                ):
+                    if not chunk:
+                        continue
+                    handle.write(chunk)
+                    transferred += len(chunk)
+                    if on_progress is not None:
+                        on_progress(transferred, total_bytes)
+            # Same directory, so this is an atomic replace.
+            os.replace(partial_path, path)
```
edwinpav: AI flag - total_bytes (from Content-Length) and transferred are never compared before os.replace

nucleus/model.py
```diff
@@ -343,3 +344,38 @@ def remove_trained_slice_ids(self, slide_ids: List[str]):
             )
 
         return response.json()
+
+    def upload_weights(self, path: str, **kwargs) -> "ModelWeights":
```
edwinpav: instead of **kwargs, can we mirror the keyword-only params explicitly, like the Benchmark wrappers do? i think this helps type checkers and IDEs as well

edwinpav: also curious, why is ModelWeights quoted here and line 369?

tests/test_model_weights.py
```diff
@@ -0,0 +1,583 @@
...
+        mock_put.return_value = _ok_put()
+        parts = _transfer_weights_to_storage(str(path), presign, 32)
+
+    assert parts is None
+    assert mock_put.call_count == 1
+    _, kwargs = mock_put.call_args
+    assert kwargs["headers"] == {"Content-Type": "application/octet-stream"}
+
+
+def test_transfer_multipart_uploads_each_part_without_headers(tmp_path):
```
edwinpav, line 145: can this also test / assert the bytes sent per part?

tests/test_model_weights.py
```diff
@@ -0,0 +1,583 @@
+"""Unit tests for model weights upload/download (no live API, no real S3)."""
```
edwinpav, line 1: Not super well versed with tests, but here's some suggested AI gap tests, up to you on what is worth testing. What I think are valid is 1, 2, 4, 6

1. GAP — no last-partial-part coverage. Every multipart test uses a file that divides evenly; a 33-byte file at part size 16 (3 parts, 1-byte tail) is untested.
2. GAP — no multipart run through upload_model_weights, so the finalize payload's parts list — order, eTag key name, presence — is never asserted on the real path.
3. GAP — no test or comment for Executor.map's cancellation behavior, which the multipart failure path silently depends on.
4. GAP — test_transfer_multipart_raises_without_etag uses a single part, so concurrent exception propagation is untested.
5. GAP — nothing asserts timeout= is passed on either transfer. Trivial to pin, prevents a regression to an unbounded hang.
6. GAP — no test for a progress callback that raises, which fires inside with progress_lock.

edwinpav (commented):
overall looks good. mostly nit comments - not too familiar with file reading and writing but the main comments are about error handling around when a non-expected amount of bytes are read/written. also for progress bar, didn't follow it super well but i've used tqdm package in python before which works well (i think it's already used within repo as well)

nucleus/__init__.py
luke-e-schaefer: id probably want to let this bubble up then catch and rethrow imo

nucleus/model_weights.py
luke-e-schaefer: dead from prototype, will remove

nucleus/model_weights.py
luke-e-schaefer: will do, thinking maybe generic was better just in case people are running this not in a ui context but want to monitor with an agent or something, but i think that might have been a little overkill on thought

nucleus/model_weights.py
luke-e-schaefer: can do, will add a 5 count backoff retry

nucleus/model_weights.py
luke-e-schaefer: bad size here i think fails on the S3 put so i dont think necessary

nucleus/model_weights.py
luke-e-schaefer: pool.map keeps the order of the input parts but the presign response doesn't guarantee those come back in order so sorted() is what makes sure the final payload is ordered

nucleus/__init__.py
```diff
@@ -1729,13 +1729,18 @@ def upload_model_weights(
             original_filename: Filename to show for the artifact. Defaults to
               the name of the file at ``path``.
             checksum_sha256: Optional SHA-256 of the artifact.
-            on_progress: Called with ``(bytes_uploaded, total_bytes)`` as the
-              upload proceeds.
+            progress: Whether to show a ``tqdm`` progress bar for the upload.
```
edwinpav, line 1732: nit: any reason why we don't always show this progress bar? just curious

nucleus/model_weights.py
```diff
@@ -86,12 +90,9 @@ class ModelWeights:
     original_filename: Optional[str] = None
     content_type: Optional[str] = None
     download_url: Optional[str] = None
-    _client: Optional["NucleusClient"] = field(repr=False, default=None)
 
     @classmethod
-    def from_json(
-        cls, payload: dict, client: Optional["NucleusClient"] = None
-    ) -> "ModelWeights":
+    def from_json(cls, payload: dict) -> "ModelWeights":
```
edwinpav, line 95: nit: This still uses quotes around ModelWeights but I saw other spots updated

nucleus/model_weights.py
```diff
@@ -115,22 +115,87 @@ def _read_part(path: str, offset: int, size: int) -> bytes:
...
 
+class _RetryableTransferError(Exception):
+    """Marks a storage failure worth retrying (network blip, 429, or 5xx).
+
+    The underlying error is attached as ``__cause__`` so it can be surfaced
+    unchanged once the retries are exhausted.
+    """
+
+
+def _is_retryable_status(status_code: int) -> bool:
```
edwinpav, line 126: nit should these functions be indented within this `_RetryableTransferError` class?

nucleus/model_weights.py
```diff
@@ -115,22 +115,87 @@ def _read_part(path: str, offset: int, size: int) -> bytes:
...
+        raise RuntimeError(message)
+
+    return _with_retries(send, reset)
+
+
+def _progress_to_bar(progress_bar: Any) -> ProgressCallback:
+    """Adapt the cumulative ``(transferred, total)`` callback to a tqdm bar."""
+
+    def update(transferred: int, total: int) -> None:
+        # pylint: disable=unused-argument
```
edwinpav: why do we need `total` if it's unused?

edwinpav (approved):

nucleus/__init__.py
edwinpav: that works - so it's caught at some point?

nucleus/model_weights.py
edwinpav: iirc `tqdm` can also show on the terminal as well so it's good when its logged

nucleus/__init__.py
luke-e-schaefer: no like it should error out...if this is failing we should know quick iykwim

luke-e-schaefer: unless you're concerned about security issues in the error stack

nucleus/__init__.py
luke-e-schaefer: might be loud for really big ones and users could just want a clean console...default is on, but maybe just in case you're wanting to pipe the stdout somewhere else

nucleus/model_weights.py
luke-e-schaefer: this is inside the ModelWeights class definition, so the name doesn't exist yet when from_json is defined so the quotes are a required forward reference (unquoting it would NameError at import)

nucleus/model_weights.py
luke-e-schaefer: CC: No — _is_retryable_status and _with_retries are intentionally module-level helpers; they operate on a status code / a callable, not on an exception instance. _RetryableTransferError is just a marker exception (body is only a docstring), so they sit next to it but aren't methods.

PR #467, opened 2026-06-25
update sdk with new adds

luke-e-schaefer:
## Summary — v0.19.0: EvaluationV2, benchmark paradigm

Brings the SDK to parity with the Nucleus EvaluationV2 UI. Evaluation creation is **benchmark-scoped only** — dataset/slice-scoped creation (an unreleased earlier iteration of this branch) was removed before release as the platform moved to the benchmark paradigm.

### Added
- **Benchmarks.** `create_benchmark()` (members from `item_ids`, `(dataset_id, ref_id)` `items` pairs, a `slice_id`, or a `dataset_id` — exactly one required; membership frozen at creation), `list_benchmarks()`, `get_benchmark()`, `update_benchmark()`, `delete_benchmark()`, `list_benchmark_items()`, plus the `Benchmark` resource (`refresh()` / `update()` / `delete()` / `items()` / `create_evaluation_v2()`) in `nucleus/benchmark.py`.
- **Benchmark evaluations.** `create_benchmark_evaluation_v2(benchmark_id, model_run_id, ...)` — every benchmark item is scored (uncovered items count as FN, keeping leaderboard scores comparable). `EvaluationV2` exposes `benchmark_id`, `rollup_groups`, `exclusion_rules`, `exclusion_stats`. No `only_items_with_predictions` / eval-scope `slice_id` (invalid in the benchmark paradigm; the server rejects the former).
- **Rollup groups.** `RollupGroup` (`class_name` + `labels`) as the primary label configuration on benchmark eval create and presets (`create/update_evaluation_v2_preset(rollup_groups=...)`, mutually exclusive with legacy `allowed_label_matches`; parsed from both key casings). Preset seeding prefers `rollup_groups`, falling back to legacy matches.
- **Exclusion rules.** `MetadataExclusionRule` / `LabelExclusionRule` / `BoxAreaExclusionRule` via `exclusion_rules` on eval create and presets.
- **Presets.** `list/create/update/delete_evaluation_v2_preset` + `EvaluationV2Preset` resource; `preset=` seeds benchmark eval creation (explicit args override).
- **Results.** `EvaluationV2.charts()` (mAP, per-class AP, confusion matrix, PR/F1 curves, TIDE, AP by size) and `examples()` (paginated TP/FP/FN; `match_type` optional) with `EvaluationV2FilterArgs` filtering; `cancel()` / `retry()`; `Dataset.evaluation_label_schema()`.
[...]

PR #472, opened 2026-08-03
Add multi-dataset model run prediction uploads

luke-e-schaefer:
## Summary — v0.20.0: multi-dataset model runs

**Stacked on #467** (`update-nuc-sdk-for-new-eval-stuff-pt1`) — review that first; this PR's diff is only the 7 files below. The SDK-visible half of the scaleapi change that gives a model run a resolved **set** of datasets instead of one declared dataset.

### The one new method

```python
dataset_b.upload_predictions_for_model_run(run_id, predictions)
```

Posts to `dataset/{dataset_id}/modelRun/{model_run_id}/uploadPredictions`, adding `dataset_b` to the run's dataset set. That is what lets a single run be scored against a benchmark whose items span several datasets — previously `create_benchmark_evaluation_v2` returned a 400 for exactly that case.

Same `update` / `asynchronous` / `batch_size` / file-batching / `trained_slice_id` arguments as `upload_predictions`, and it runs the same duplicate-id check.

### What is deliberately *not* changed

`Dataset.upload_predictions` still cannot widen a run, and that's the point. It identifies the run by `(dataset, model)`, so it finds the run already on this dataset or creates a new one. Widening got its own endpoint rather than loosening this one — mirroring the server, where the existing route kept its never-widen contract exactly as it was.

`ModelRun.predict` also stays on the old route. It's deprecated, and switching it would silently turn a stale `dataset_id` passed to `get_model_run()` into a widening upload. Its docstring now says it fails for multi-dataset runs and points at the new method.

### Routing

`PredictionUploader` now accepts `dataset_id` together with `model_run_id` — previously an assertion rejected the pair — and picks the endpoint from which identifiers are present:

| Arguments | Route | Can widen? |
| --- | --- | --- |
| `dataset_id` + `model_run_id` | `dataset/{ds}/modelRun/{run}/uploadPredictions` | **Yes** |
| `dataset_id` + `model_id` | `dataset/{ds}/model/{model}/uploadPredictions` | No |
| `model_run_id` alone | `modelRun/{run}/predict` | No (deprecated) |

The two pre-existing forms route exactly as before.

### Access, and the sharp edge

Widening requires write on this dataset **and** on every dataset the run already covers. That's stricter than it looks necessary, for a reason worth knowing: a run is visible only to users who can read *all* of its datasets, so adding a dataset to a run can remove that run from a collaborator's view. The set also only ever grows — a later upload never drops a dataset, so it can't widen who can read the run.

### Docstring corrections

Both are wrong as of the server change, not merely incomplete:

- `create_benchmark_evaluation_v2` / `Benchmark.create_evaluation_v2` said the run's predictions "must cover items from the benchmark's datasets". The server enforced that with a 400; it no longer does. Coverage may be partial or empty, and uncovered members score as false negatives.
- `ModelRun.predict` — see above.

### Tests

`tests/test_multi_dataset_model_runs.py` — 9 mock-based tests pinning all three routes (the whole difference between them *is* the route), the async route, `trained_slice_id` forwarding, and that the new entry point runs the duplicate-id check rather than bypassing it.

Ran locally: **68 passed** across the new file plus the benchmark / eval-v2 / preset / leaderboard suites. `black` and `isort` clean on the changed files. Diff is purely additive — no reformatting churn from a newer local `black`.

> **Server dependency:** the new route ships with the multi-dataset model-run work in scaleapi ([#154100](https://github.com/scaleapi/scaleapi/pull/154100)). Unit tests pass regardless; live calls 404 until that deploys.

<h3>Greptile Summary</h3>

Adds multi-dataset model-run prediction uploads while preserving existing upload behavior.
- Adds `Dataset.upload_predictions_for_model_run` with synchronous and asynchronous widening routes.
- Refactors shared prediction-upload validation and dispatch into `_upload_predictions`.
- Keeps deprecated `ModelRun.predict` on its existing endpoint and documents its multi-dataset limitation.
- Updates benchmark-evaluation documentation, package version, changelog, and route-focused tests.

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| nucleus/dataset.py | Adds the public multi-dataset upload entry point and extracts common synchronous/asynchronous upload handling. |
| nucleus/annotation_uploader.py | Extends prediction uploader routing to support explicit dataset-and-run uploads while validating identifier combinations. |
| nucleus/model_run.py | Preserves the deprecated prediction route through an explicit override and documents its multi-dataset limitation. |
| tests/test_multi_dataset_model_runs.py | Covers identifier validation, all upload routes, asynchronous routing, trained-slice forwarding, and duplicate-ID rejection. |
| nucleus/__init__.py | Corrects benchmark-evaluation documentation to describe partial or empty model-run coverage. |

</details>

```mermaid
flowchart TD
    A[Prediction upload] --> B{Identifiers supplied}
    B -->|dataset_id + model_run_id| C[dataset/:dataset/modelRun/:run/uploadPredictions]
    B -->|dataset_id + model_id| D[dataset/:dataset/model/:model/uploadPredictions]
    B -->|Deprecated ModelRun.predict| E[modelRun/:run/predict]
    C --> F[Existing run may gain dataset]
    D --> G[Resolve or create run for dataset and model]
    E --> H[Retain legacy single-dataset behavior]
```
</details>

<sub>Reviews (7): Last reviewed commit: ["updatte"](https://github.com/scaleapi/nucleus-python-client/commit/458e26f455d0287c3199eb73a2f23b4fa0d5f173) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49715310)</sub>


nucleus/annotation_uploader.py
```diff
@@ -247,8 +270,13 @@ def __init__(
         super().__init__(dataset_id, client)
         self._client = client
         self.trained_slice_id = trained_slice_id
-        if model_run_id is not None:
-            assert model_id is None and dataset_id is None
+        if model_run_id is not None and dataset_id is not None:
+            assert (
```
jaypsiri: I think this would be a good opportunity to get rid of the `assert` in the logic here - it could make this be something like:

```
if model_id is None:
    raise Error("Pass either model_id or model_run_id, not both.")
```

nucleus/annotation_uploader.py
```diff
@@ -247,8 +270,13 @@ def __init__(
...
+                model_id is None
+            ), "Pass either model_id or model_run_id, not both."
+            self._route = f"dataset/{dataset_id}/modelRun/{model_run_id}/uploadPredictions"
+        elif model_run_id is not None:
+            assert model_id is None
```
jaypsiri: Same thing here

nucleus/annotation_uploader.py
```diff
@@ -247,8 +270,13 @@ def __init__(
...
             self._route = f"modelRun/{model_run_id}/predict"
         else:
             assert (
```
jaypsiri: not you, but if you fix the first two can we also change this `assert` to make it a `raise` instead for succinctness

nucleus/annotation_uploader.py
```diff
@@ -236,6 +236,29 @@ def check_for_duplicate_ids(self, annotations: Iterable[Annotation]):
...
+        every dataset the run already covers, not just this one, because a run
+        is only visible to users who can read all of its datasets.
+
+    ``dataset_id`` + ``model_id``
+        ``dataset/{dataset_id}/model/{model_id}/uploadPredictions``. Resolves
+        (or creates) the run for that model on that dataset. Cannot widen an
+        existing run: passing a ``model_run_id`` belonging to a different
+        dataset is rejected server-side.
+
+    ``model_run_id`` alone
```
jaypsiri: I might a bit confused by this docstring - so it uploads to three endpoints, but the last one is deprecated and could possibly be rejected by the server? should we get rid of the last endpoint (and then it's just two)? or is that something to be done in a later PR

nucleus/__init__.py
```diff
@@ -1265,10 +1265,16 @@ def create_benchmark_evaluation_v2(
         background — call :meth:`EvaluationV2.wait_for_completion`, then
         :meth:`EvaluationV2.charts` or :meth:`EvaluationV2.examples`.
 
+        The benchmark may span datasets the model run has no predictions in at
+        all. Those members are scored as false negatives like any other
```
jaypsiri, line 1328: question for myself - if the model didn't make a prediction, why is it scored as a false negative instead of discarded? wouldn't that impact the recall/F1 score/other metrics in a way that doesn't reflect reality?

nucleus/dataset.py
```diff
@@ -2070,6 +2070,97 @@ def upload_predictions(
             trained_slice_id=trained_slice_id,
         )
 
+    def upload_predictions_for_model_run(
```
jaypsiri, line 2058: can you add a type hint for the return type here

jaypsiri, line 2058: it looks like this duplicates a lot of the code from `upload_predictions` - it might be worth extracting common stuff into a shared helper?

jaypsiri (commented):
had a few comments/questions!

nucleus/__init__.py
luke-e-schaefer: Its more semantics; every benchmark item is scored to stay consistent basically so a run with no prediction on an item takes an FN to keep leaderboard scores comparable across runs -- i.e a run that skips items shouldn't out-rank/out-perform one that attempts them. Discarding uncovered items would inflate precision/recall against the benchmark; users can run custom evals against subsets of a benchmark, but it has to be specified

nucleus/annotation_uploader.py
luke-e-schaefer: yeah I can remove thats probably better

nucleus/model_run.py
```diff
@@ -166,8 +172,13 @@ def predict(
                 "predictions_ignored": int,
             }
         """
+        # Deprecated route: PredictionUploader no longer derives it from ids, so
```
jaypsiri: maybe get rid of this comment it is a bit extraneous

nucleus/model_run.py
```diff
@@ -137,6 +137,12 @@ def predict(
     ) -> Union[dict, AsyncJob]:
         """Uploads model outputs as predictions for a model_run.
 
+        Deprecated along with the rest of this class. The target dataset is
+        inferred from the run rather than named, so this fails for a run that
+        spans more than one dataset — there is no single dataset to infer. Use
+        :meth:`Dataset.upload_predictions_for_model_run` instead, which takes
+        both ids explicitly.
```
jaypsiri: really nit: if we're going to deprecate/delete this later, then leaving this comment here makes sense - but for now can we update it to name the two ids its taking it so its clear for client-facing usage

jaypsiri (approved):
a few nits on comment stuff but other than this lgtm!

PR #473, opened 2026-08-14
feat(nucleus): Expose dataset_item_id on exported items, annotations and predictions

luke-e-schaefer:
Batch exports only ever returned `reference_id` on the exported item, so keying an exported prediction back to a Nucleus dataset item meant a second lookup.

This is an **inconsistency fix, not a new field**: the single-item endpoints (`/dataset/:id/loc`, `/refloc`, `/iloc`) already return `dataset_item_id`, and the public API docs (`ApiDocsPage/slices/batch-export.md`) already document it in the batch-export response. The batch export was the odd one out.

Pairs with scaleapi PR: `luke/nucleus-export-dataset-item-id`.

## What changed

`dataset_item_id` now appears everywhere `reference_id` already does:

- `DatasetItem` gains the field, wired through `from_json` — the single deserialization entry point, so every SDK path that returns an item picks it up, not just exports.
- Every `Annotation` / `Prediction` subclass gains it (`box`, `line`, `polygon`, `keypoints`, `cuboid`, `category`, `multicategory`, `segmentation`). `convert_export_payload` stamps it down from the item exactly as it already does for `reference_id`.
- The scene/video export docstring documents it on each track frame.

## Read-only semantics

The field is server-assigned, and mirrors how `DatasetItem.phash` was done:

| | behaviour |
|---|---|
| `from_json` | populated |
| locally constructed | `None` |
| `__eq__` | excluded (`compare=False`) |
| `to_payload` | absent — uploads unchanged |

The `__eq__` exclusion matters: without it, every existing test that compares a locally-built annotation against its round-tripped twin would start failing.

Exports from a backend that doesn't return the field leave it `None` rather than throwing, so this SDK version is safe against an un-upgraded backend.

## Verification

The test suite requires live API keys (`conftest.py` hard-asserts on `NUCLEUS_PYTEST_API_KEY`), so it could not be run locally. Verified offline instead by driving `convert_export_payload` directly:

- every geometry type on both the annotation and prediction paths carries the id
- a local object still compares equal to its round-tripped self
- `to_payload` output is unchanged
- an item payload with no `dataset_item_id` yields `None` everywhere and does not raise

Formatted with the repo-pinned black 23.12.1 (a newer local black introduced unrelated reformatting, which was reverted).

<h3>Greptile Summary</h3>

This PR exposes the server-assigned `dataset_item_id` throughout deserialized dataset items, annotations, and predictions without adding it to upload payloads or equality comparisons.
- Adds read-only `dataset_item_id` fields to dataset-item and object models.
- Propagates the item identifier through batch and single-item response conversion.
- Adds offline coverage for annotation and prediction propagation, backward compatibility with older backends, and serialization behavior.
- Documents the identifier in scene/video export frames and the changelog.

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| nucleus/annotation.py | Adds a keyword-only, read-only dataset item identifier to all annotation models and populates it during deserialization. |
| nucleus/prediction.py | Extends prediction constructors and deserializers to preserve the server-returned dataset item identifier without changing existing positional parameters. |
| nucleus/dataset_item.py | Adds the non-comparing dataset item identifier to deserialized `DatasetItem` objects while keeping upload payloads unchanged. |
| nucleus/utils.py | Propagates the item-level identifier into converted annotations and predictions across batch and single-item response paths. |
| tests/test_utils.py | Adds offline tests for identifier propagation, omitted backend values, equality, keyword-only construction, and unchanged upload serialization. |
| nucleus/dataset.py | Documents `dataset_item_id` on scene and video track frames. |
| CHANGELOG.md | Documents the new SDK exposure and its read-only compatibility semantics. |

</details>

```mermaid
flowchart LR
    A[Backend export response] --> B[DatasetItem.from_json]
    A --> C[Response conversion]
    C --> D[Annotation.from_json]
    C --> E[Prediction.from_json]
    B --> F[DatasetItem.dataset_item_id]
    D --> G[Annotation.dataset_item_id]
    E --> H[Prediction.dataset_item_id]
    F -. omitted .-> I[to_payload]
    G -. omitted .-> I
    H -. omitted .-> I
```
</details>

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;master&#39; into lukeschaefer/..."](https://github.com/scaleapi/nucleus-python-client/commit/1a51b948364e840221ed36a0e92846634b97f76c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54025623)</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to the [Nucleus Python Client](https://github.com/scaleapi/n
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.20.1](https://github.com/scaleapi/nucleus-python-client/releases/tag/v0.20.1) - 2026-08-14
+## [0.20.1](https://github.com/scaleapi/nucleus-python-client/releases/tag/v0.20.1) - 2026-08-18
 
 ### Added
 - **`dataset_item_id` on exported items and objects.** Batch exports now carry the Nucleus-internal dataset item id (`di_*`) everywhere `reference_id` already appeared: on `DatasetItem`, and on every exported annotation and prediction (`box`, `line`, `polygon`, `keypoints`, `cuboid`, `category`, `multicategory`, `segmentation`). Video/scene exports carry it on each track frame. Previously only `reference_id` was returned, so keying predictions back to items required a second lookup.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to the [Nucleus Python Client](https://github.com/scaleapi/n
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.20.1](https://github.com/scaleapi/nucleus-python-client/releases/tag/v0.20.1) - 2026-08-14
+
+### Added
+- **`dataset_item_id` on exported items and objects.** Batch exports now carry the Nucleus-internal dataset item id (`di_*`) everywhere `reference_id` already appeared: on `DatasetItem`, and on every exported annotation and prediction (`box`, `line`, `polygon`, `keypoints`, `cuboid`, `category`, `multicategory`, `segmentation`). Video/scene exports carry it on each track frame. Previously only `reference_id` was returned, so keying predictions back to items required a second lookup.
+
+  The field is server-assigned and read-only: it is populated by `from_json`, left `None` on objects you construct locally, excluded from `__eq__`, and never sent in `to_payload`. Exports from an older backend that does not return it simply leave it `None`.
+
 ## [0.20.0](https://github.com/scaleapi/nucleus-python-client/releases/tag/v0.20.0) - 2026-08-11
 
 ### Added
--- a/nucleus/annotation.py
+++ b/nucleus/annotation.py
@@ -51,6 +51,20 @@ from .constants import (
 # pylint: disable=C0302
 
 
+def _server_assigned_id_field():
+    """Dataclass field for the Nucleus-internal dataset item id (``di_*``) of the
+    item this object sits on.
+
+    Server-assigned and read-only: populated on objects returned by the API,
+    ``None`` on ones you construct locally to upload, and never sent in
+    ``to_payload``. Excluded from ``__eq__`` so a locally-built object still
+    compares equal to its round-tripped self (same reason as
+    ``DatasetItem.phash``).
+    """
+    # pylint: disable=invalid-field-call  # returned into a dataclass body below
+    return field(default=None, repr=False, compare=False)
+
+
 class Annotation:
     """Internal base class, not to be used directly.
 
@@ -59,8 +73,8 @@ class Annotation:
     """
 
     reference_id: str
-    # Set on every subclass so callers can read it off the base type. See the note on
-    # BoxAnnotation for why it is server-assigned and excluded from equality.
+    # Set on every subclass (via ``_server_assigned_id_field``) so callers can read
+    # it off the base type; server-assigned and excluded from equality.
     dataset_item_id: Optional[str]
 
     @classmethod
@@ -164,14 +178,7 @@ class BoxAnnotation(Annotation):  # pylint: disable=R0902
     embedding_vector: Optional[list] = None
     track_reference_id: Optional[str] = None
     _task_id: Optional[str] = field(default=None, repr=False)
-    # Nucleus-internal dataset item id (``di_*``) of the item this object sits on.
-    # Server-assigned and read-only: populated on objects returned by the API, ``None``
-    # on ones you construct locally to upload, and never sent in ``to_payload``.
-    # Excluded from ``__eq__`` so a locally-built object still compares equal to its
-    # round-tripped self (same reason as ``DatasetItem.phash``).
-    dataset_item_id: Optional[str] = field(
-        default=None, repr=False, compare=False
-    )
+    dataset_item_id: Optional[str] = _server_assigned_id_field()
 
     def __post_init__(self):
         self.metadata = self.metadata if self.metadata else {}
@@ -294,14 +301,7 @@ class LineAnnotation(Annotation):
     metadata: Optional[Dict] = None
     track_reference_id: Optional[str] = None
     _task_id: Optional[str] = field(default=None, repr=False)
-    # Nucleus-internal dataset item id (``di_*``) of the item this object sits on.
-    # Server-assigned and read-only: populated on objects returned by the API, ``None``
-    # on ones you construct locally to upload, and never sent in ``to_payload``.
-    # Excluded from ``__eq__`` so a locally-built object still compares equal to its
-    # round-tripped self (same reason as ``DatasetItem.phash``).
-    dataset_item_id: Optional[str] = field(
-        default=None, repr=False, compare=False
-    )
+    dataset_item_id: Optional[str] = _server_assigned_id_field()
 
     def __post_init__(self):
         self.metadata = self.metadata if self.metadata else {}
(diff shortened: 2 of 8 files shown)


edwinpav:
👀

nucleus/prediction.py
```diff
@@ -129,6 +131,7 @@ def from_json(cls, payload: dict):
                 for ann in payload.get(ANNOTATIONS_KEY, [])
             ],
             reference_id=payload[REFERENCE_ID_KEY],
+            dataset_item_id=payload.get(DATASET_ITEM_ID_KEY, None),
```
edwinpav: does the docs comment above (lines 93-122) need to be updated to include this key? Also do we have to fall back to `None` here, doesn't seem like we do that in the `from_json` in `dataset_item.py`

nucleus/prediction.py
```diff
@@ -189,6 +192,7 @@ def __init__(
         class_pdf: Optional[Dict] = None,
         embedding_vector: Optional[list] = None,
         track_reference_id: Optional[str] = None,
+        dataset_item_id: Optional[str] = None,
```
edwinpav, line 196: is comment update also needed here? is it not needed because it's set server side? is there somewhere on this client-side sdk that prevents a user from giving it a field themselves?

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to the [Nucleus Python Client](https://github.com/scaleapi/n
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.20.1](https://github.com/scaleapi/nucleus-python-client/releases/tag/v0.20.1) - 2026-08-14
```
edwinpav: very nit: update release date

nucleus/annotation.py
```diff
@@ -160,6 +164,14 @@ class BoxAnnotation(Annotation):  # pylint: disable=R0902
...
     track_reference_id: Optional[str] = None
     _task_id: Optional[str] = field(default=None, repr=False)
+    # Nucleus-internal dataset item id (``di_*``) of the item this object sits on.
+    # Server-assigned and read-only: populated on objects returned by the API, ``None``
+    # on ones you construct locally to upload, and never sent in ``to_payload``.
+    # Excluded from ``__eq__`` so a locally-built object still compares equal to its
+    # round-tripped self (same reason as ``DatasetItem.phash``).
+    dataset_item_id: Optional[str] = field(
+        default=None, repr=False, compare=False
+    )
```
edwinpav: this is repeated a bunch with comment throughout this file. Should we create some private helper function and reuse that?

Like this:

```
def _server_assigned_id_field():
    """<Comment here>"""
    return field(default=None, repr=False, compare=False)
```

nucleus/dataset.py
```diff
@@ -1584,6 +1584,7 @@ def scene_and_annotation_generator(
                                 "width": int,
                                 "height": int,
                                 "key": str, # frame key
+                                "dataset_item_id": str, # id of the frame's dataset item
```
edwinpav: nit: should comment be a bit more specific to not confuse user (`# internally generated id of the frame's dataset item`)

edwinpav: do we also need to update comments for the other export functions like `items_and_annotations`, `items_and_annotation_generator` and stuff within `slice.py`?

nucleus/dataset_item.py
```diff
@@ -187,6 +193,7 @@ def from_json(cls, payload: dict):
             reference_id=payload.get(REFERENCE_ID_KEY),
             metadata=payload.get(METADATA_KEY, {}),
             phash=payload.get(PHASH_KEY),
+            dataset_item_id=payload.get(DATASET_ITEM_ID_KEY),
```
edwinpav, line 196: should this also fall back to `None`?

nucleus/prediction.py
edwinpav: More in-depth AI comment about my last question above:

`dataset_item_id` is now a positionally-reachable public constructor param on all 8 prediction classes (and via the dataclass on all 9 annotations) for a field documented as server-assigned and read-only. `BoxPrediction("l", 0, 0, 1, 1, "r", 0.9, "a1", {}, {}, None, None, "di_spoofed")` is accepted and then silently dropped by to_payload(). Making it keyword-only (*, dataset_item_id: Optional[str] = None) would encode "not part of the upload contract" in the signature. Cheap and non-breaking — nobody can be passing it positionally yet.

"is there somewhere on this client-side sdk that prevents a user from giving it a field themselves?" — No, nothing prevents it. Verified:

BoxPrediction('l',0,0,1,1,'r',0.9,'a1',{},{},None,None,'di_spoofed')
accepted positionally -> di_spoofed
in to_payload -> False

It's accepted positionally, sets .dataset_item_id = "di_spoofed" on the object, and to_payload() then silently discards it. No validation, no warning, no error — the user gets an object whose attribute disagrees with what was uploaded. That's the gap that makes the keyword-only suggestion (*, dataset_item_id: Optional[str] = None) worth pushing on: it at least removes the positional footgun and signals "not part of the upload contract" in the signature itself.

nucleus/utils.py
```diff
@@ -239,10 +240,15 @@ def convert_export_payload(api_payload, has_predictions: bool = False):
     for row in api_payload:
         return_payload_row = {}
         return_payload_row[ITEM_KEY] = DatasetItem.from_json(row[ITEM_KEY])
+        # The backend returns reference_id and dataset_item_id on the item, not on each
```
edwinpav, line 248: Does `format_dataset_item_response` also need to be updated?

edwinpav (commented):
Couple comments, aside from that, can / should any unit tests that test these existing functions be made or updated?

nucleus/dataset_item.py
luke-e-schaefer: yeah it already does i think

nucleus/utils.py
luke-e-schaefer: CC: " Not needed — items_and_annotations, items_and_annotation_generator (in both dataset.py and slice.py) all route through convert_export_payload, which already stamps the id, so they pick it up for free. Only the scene export documents per-frame fields explicitly (it returns raw dicts), which is why that's the one docstring touched."

nucleus/prediction.py
luke-e-schaefer: idt the other docstrings list dataset_item_id since it's server-assigned/read-only and not a user-supplied param  (idt we document _task_id)

nucleus/dataset.py
luke-e-schaefer: CC: Not needed — those all route through convert_export_payload, which already stamps the id, so they pick it up for free. Only the scene export documents per-frame fields explicitly (it returns raw dicts), which is why that's the one docstring touched.

nucleus/annotation.py
```diff
@@ -51,6 +51,24 @@
 # pylint: disable=C0302
 
 
+def _server_assigned_id_field():
+    """Dataclass field for the Nucleus-internal dataset item id (``di_*``) of the
```
edwinpav, line 55: can you double check that this function / comment won't show up on the public docs since this should be just an internal function?

nucleus/prediction.py
edwinpav: ah yea makes sense if the doc is for params and not output

nucleus/utils.py
```diff
@@ -224,7 +231,9 @@ def format_scale_task_info_response(response: dict) -> Union[Dict, List[Dict]]:
     return ret
 
 
-# pylint: disable=too-many-branches
+# One branch/statement per annotation type — stamping reference_id and
+# dataset_item_id onto each pushed it just over pylint's default ceilings.
```
edwinpav: this can be removed, don't need comment explaining the pylint disable stuff

edwinpav (approved):

nucleus/annotation.py
luke-e-schaefer: should be hidden from docs

